PR #2, opened 2022-06-02
Use `unix` metadata extensions, rather than linux-specific ones

Dr-Emann:
This at least makes it compile, and pass what tests there are on my macbook.

### DIFF
--- a/src/backend/ignore.rs
+++ b/src/backend/ignore.rs
@@ -1,5 +1,5 @@
 use std::fs::{read_link, File};
-use std::os::linux::fs::MetadataExt;
+use std::os::unix::fs::MetadataExt;
 use std::path::{Path, PathBuf};
 
 use anyhow::Result;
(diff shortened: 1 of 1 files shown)


aawsome:
So far there are not many tests, unfortunately.
I myself only have linux machines at hand to run tests but I'm very excited to get rustic working on other OSes, too.

Did you try a sequence of `rustic-rs init`, `rust-rs backup` with you branch on a macbook and took a look at some saved trees? 

Cating saved trees can be done by `rustic-rs cat tree <SNAPSHOT-ID> <PATH>`
where `SNAPSHOT-ID` can be the (unique) start of an id, like a1bc5f or even `latest`.
In my example I would e.g. run `rustic-rs -r /tmp/repo cat tree latest /home/alex-dev/rust/rustic-rs/src/commands`

Dr-Emann:
It looks like it works to me, yep

aawsome:
@Dr-Emann: Thanks a lot for proposing this PR!

I manually checked that this PR also doesn't change anything on linux.
So, looks good to me!

Which targets are relevant for supporting macbooks? Are these two sufficient?
aarch64-apple-darwin
x86_64-apple-darwin

aawsome (approved):
LGTM

aawsome:
I also added tests on MacOS on the CI pipeline, see #8

PR #4, opened 2022-06-08
Change wording for `check` to match restic

dimejo:


aawsome:
LGTM! Thanks for contributing!

PR #18, opened 2022-06-13
add rust audit

aawsome:
currently failing due to security issue in chrono (fixed in unreleased version 0.4.20), see https://github.com/chronotope/chrono/issues/602

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,32 @@
+name: Security audit
+
+on:
+  schedule:
+    # Runs at 00:00 UTC everyday
+    - cron: '0 0 * * *'
+  push:
+    paths:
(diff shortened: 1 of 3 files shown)


dimejo:
I think you could skip this check by adding `args: --ignore RUSTSEC-2020-0071`.

aawsome:
Seems like `actions-rs/audit-check@v1` doesn't support adding custom args, unfortunately. I suggest waiting for chrono 0.4.

aawsome:
Now chrono 0.4.20 is out. But `cargo audit` still  claims the dependency on `time` which doesn't seem to be removed. So, we wait a bit more for what's happening...

BTW: The other finding is a warning about `term` being unmaintained. This should be fixed once #42 is addressed.

PR #61, opened 2022-07-13
Limit concurrent operations

aawsome:
Limit the number of concurrent reads/writes.
This is needed for some remote backends which refuse too many connections.

Also reduces memory requirements.

### DIFF
--- a/src/backend/decrypt.rs
+++ b/src/backend/decrypt.rs
@@ -2,7 +2,7 @@ use std::fs::File;
 
 use anyhow::{bail, Result};
 use async_trait::async_trait;
-use futures::{stream::FuturesUnordered, TryStreamExt};
+use futures::{stream, stream::FuturesUnordered, StreamExt};
 use indicatif::ProgressBar;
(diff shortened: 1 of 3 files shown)


borkd:
@aawsome - is hardcoding limits to cater to the most handicapped backends the best option? Please consider allowing such limits to be set via cli flag.

aawsome:
@borkd Thanks for your comment. When changing this, I first thought about making this public options. Plus define reasonable defaults for each backend type (local vs. remote backend).

I did some test though with local repos and didn't observe performance regression in my cases but did observe a decrease in memory usage. This could change with very powerful machines ans huge repos - in this case 
Altogether, I decided to run fast first here and first make sure remote backends do work. I'll add an issue to expose the parameters to users. If you can give some example where this finetuning is essential, I'll put higher priority onto this topic.

borkd:
As for my use case - still experimenting and don't have a lot of info on hand. Backend storage performance profile seems to be key limiting factor and I'd like to see what would it take to effectively balance work across more than 20 cores during "check" after the initial index is loaded

PR #70, opened 2022-07-14
prune: Add option --instant-delete

aawsome:
Add the option `--instant-delete` to `prune`. With this option, `rustic prune` uses no two-phase pruning but removes unneeded pack files immediately.
This is identically to the behavior of restic except that no locks are created. 
USE THIS OPTION WITH CARE!

The option is interesting for exclusive pruning, especially in combination with #7.

### DIFF
--- a/src/commands/prune.rs
+++ b/src/commands/prune.rs
@@ -34,6 +34,11 @@ pub(super) struct Opts {
     #[clap(long, value_name = "DURATION", default_value = "23h")]
     keep_delete: humantime::Duration,
 
+    /// delete files immediately instead of marking them. This also removes all already marked files.
+    /// WARNING: Only use if you are sure the repository is not accessed by parallel processes!
+    #[clap(long)]
+    instant_delete: bool,
+
     /// minimum duration (e.g. 90d) to keep packs before repacking or removing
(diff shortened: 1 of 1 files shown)


aawsome:
You have to use `forget --prune --instant-delete`. Without `--prune`, forget does not prune anything, only removes snapshot files.

Instant-delete does exactly what restic is doing except creating locks.
So the question is mainly, what rustic is doing when you don't specify `instant-delete`: It uses a two-phase deletion, i.e. packs-to-remove are not immediately removed but first marked for some time, see https://github.com/rustic-rs/rustic/blob/main/FAQ.md

This two-phase deletion is needed for rustic to work lock-free: If a `backup` runs parallel to a `prune` run (or `forget --prune`), it could be that `prune` decides that some blobs can be removed, but the parallel `backup` uses these blobs for the newly generated snapshot.

The time to hold marked packs should be long enough to guarantee that a possibly parallel backup run has finished in between. It can be set by the `--keep-delete` option and defaults to 23 hours. In any case, packs will be kept marked and only deleted by the next `prune` run.

The `--instant-delete` now circumvents this protection and instantly removes the packs-to-delete instead of marking them. So it is only advised if you *KNOW* that no parallel access to your repo is made.

PR #90, opened 2022-07-29
restore: use existing fileparts

aawsome:
With this PR, `restore` checks the contents of possibly existing files-to-restore (if the size also matches).
If contents match to blobs which would be restore, they are used to:
- omit the content restore for these blobs (as they are already correct)
- use that data to restore other files which need the same blobs

This can extremely speed-up the restore and minimizes read-access to the repository.

Moreover, now aborted restores can be efficiently resumed. 
It also allows to verify if a backup matches existing data by simply running `restore -n` and checking that `restore` reports all file contents are fine.

Note that this PR may slow down the restore in some edge cases compared to the current behavior:
- if reading from the restore destination is very slow
- if many/big  files are read but it turns out, that all blobs need to be downloaded

To remedy these possibly negative effects, there is a simple workaround: Restore to an empty (or non-existing) destination directory.

### DIFF
--- a/src/backend/local.rs
+++ b/src/backend/local.rs
@@ -289,6 +289,29 @@ impl LocalBackend {
         Ok(())
     }
 
+    pub fn read_at(&self, item: impl AsRef<Path>, offset: u64, length: u64) -> Result<Vec<u8>> {
+        let filename = self.path.join(item);
+        let mut file = File::open(&filename)?;
+        file.seek(SeekFrom::Start(offset))?;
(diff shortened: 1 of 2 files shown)


aawsome:
> Just to confirm, in addition to allowing for resuming interrupted transfers, this will also skip files that already exist in the destination?

Correct. rustic will not save any data in a file which already has the desired content.
It will however set metadata like permissions for all files - even if this doesn't change anything.

PR #113, opened 2022-08-07
Don't use temporary files; add retries to rest/rclone backend

### LINKED ISSUE #86: retry temporary failures in rest/rclone backend
use the [backoff crate](https://crates.io/crates/backoff)

aawsome:
This PR does a bit of refactoring in the backend:
- uses `Bytes` instead of `Vec<u8>` (this allows to cheaply `clone()` data and saves memory)
- Save unfinished pack files in-memory (in the Bytes struct) instead of using a temporary file. This should make the packing even faster. However, it will increase the memory usage by the size of the pack file. As most modern distributions already use a tmpfs for `/tmp`, the overall memory consumption is however not changed in that case.
- Allows to set options for backends.
- Add retries for remote backends (actually the two upper changes are prerequisites for this change)

TODOs:
- [x] save unfinished pack file in memory instead of in temp file
- [ ] Add warning about pack sizes
- [x] Use backoff crate
- [x] Disable backoff for warm-up

closes #86

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1463,6 +1463,7 @@ dependencies = [
  "backoff",
  "base64",
  "binrw",
+ "bytes",
  "bytesize",
  "cachedir",
  "cdc",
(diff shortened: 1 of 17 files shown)


aawsome:
I'll leave the docu point open and merge this PR.

PR #194, opened 2022-09-12
ci: strip via config, not manually

istudyatuni:


aawsome:
Thanks for the PR! Are you sure this also works on the cross-compiled targets? Unfortunately, the CI doesn't run the builds as there the github credentials are used...

istudyatuni:
No 🥲 https://github.com/istudyatuni/rustic/actions/runs/3038707602/jobs/4892754814

```bash
[cross] warning: `cross` does not provide a Docker image for target aarch64-apple-darwin, specify a custom image in `Cross.toml`.
[cross] note: Falling back to `cargo` on the host.
  # --- snip ---
error: could not execute process `target/aarch64-apple-darwin/release/rustic` (never executed)
Error: could not execute process `target/aarch64-apple-darwin/release/rustic` (never executed)
Caused by:
  Bad CPU type in executable (os error 86)
```

aawsome:
Sorry, just saw that you corrected your text.
I would say, if the CI tests work, we'll merge this. The binaries will then be available as beta builds.

aawsome (approved):
LGTM

aawsome:
I just checked the binary sizes of the beta builds - seems to work!
Thanks for simplifying the CI!

PR #195, opened 2022-09-12
feat: shell completions

### LINKED ISSUE #193: Make shell completions
[Documentation](https://docs.rs/clap_complete/latest/clap_complete/generator/fn.generate_to.html) with example how to generate completions at compile-time

_I can create a PR_

istudyatuni:
Close #193 

~~Depends on~~ #194

Not sure where is better place to write docs about installing completions. Also that this way for generating is good


src/commands/mod.rs
```diff
@@ -172,7 +177,21 @@ enum Command {
...
 
+#[cfg_attr(feature = "completions", allow(unused))]
 pub async fn execute() -> Result<()> {
+    #[cfg(feature = "completions")]
+    {
+        // Generate completions if running with "completions" feature enabled
+        //
+        // A bit hacky. Including source code of module in build.rs not work
+        // because there are more one file with clap::Command declaration
+        //
```
aawsome, line 188: I don't know if clap with subcommands implementing `clap::Command` is the right way. If not, we should of course change this.

Actually, I like the "hack" with adding the feature and then using the built binary to produce the completions. What about adding a standard subcommand `generate` which is only activated when activating the feature `completions`? Then the code to actually generate the completions could go into `src/commands/generate.rs` and if anyone wants to distribute a version which is able to generate subcommands can compile with the feature selected.

.github/workflows/release.yaml
```diff
@@ -76,31 +76,23 @@ jobs:
...
-            GCC_PREFIX="aarch64-linux-gnu-"
-          else
-            GCC_PREFIX=""
-          fi
-          "$GCC_PREFIX"strip $BINARY_NAME
-
           ########## create tar.gz ##########
           RELEASE_NAME=rustic-${GITHUB_REF_NAME}-${{ matrix.job.target}}
-          tar czvf $RELEASE_NAME.tar.gz $BINARY_NAME
+          tar czvf $RELEASE_NAME.tar.gz rustic${{ matrix.job.binary-postfix }} -C ../../../target ./completions
```
aawsome, line 95: I'm a bit unsure if the completions should be released withing the tar.gz containing the binary. I would tend to creating a `completions.tar.gz` and release this as extra file.

src/commands/mod.rs
istudyatuni: Thought about command, but name `completions`, like in [poetry](https://python-poetry.org/docs/#enable-tab-completion-for-bash-fish-or-zsh)

aawsome:
@istudyatuni Are you still working on this PR? I would love to add this feature to the next release which I'm planning to release within the next days.
Also, please give me a hint if you need any kind of support!

istudyatuni:
Yeah, sorry, studying now a lot. Will return to this in 1-3 days (I hope). Also question - maybe do not hide completions behind feature, and provide this by default? Like in poetry

aawsome:
No worries and no need to hurry. I'm just trying to get things closed which can be closed :wink:

> Also question - maybe do not hide completions behind feature, and provide this by default? Like in poetry

Yes, this is for sure an option.

istudyatuni:
Done

.github/workflows/check-and-lint.yaml
```diff
@@ -22,6 +22,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: check
+          args: --all-features
```
aawsome, line 25: As this is no longer needed, maybe remove this line?

aawsome (commented):
Thanks, looks pretty good to me.
Can you squash your commits into 1 commit?

.github/workflows/check-and-lint.yaml
istudyatuni: I saw the `all-features` flag in clippy in the same file, and thought to not remove this. But I can, no problem.

istudyatuni:
About squashing - maybe make a squash merge?

.github/workflows/check-and-lint.yaml
aawsome: we can also leave it as it is..

aawsome (approved):
LGTM

PR #268, opened 2022-10-20
WIP: remove async code and parallelize using threads

aawsome:
I've been racking my brain for the last couple of weeks how to parallelize and tweak performance of rustic. I tried quite some optimizations just to find out they didn't work at all or traded a slightly improved parallelization for a huge increase in memory requirement - or made rustic even slower..

Recapturing, I think one of the point is that within rustic, we want to have a lot of heavy cpu work parallelized which is not the typical setting async code is for.

So I decided to give it a try and convert rustic back to sync code in order to use "traditional" parallelization methods...

This is WIP, but IMO it looks already pretty good and most commands are much better parallelized than before... Feedback at any time heavily appreciated!

Open points:
- [x] fully remove the async code
- [ ] identify all places where to parallelize and do so
- [ ] identify code simplifications due to the removal of async code

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -357,6 +357,40 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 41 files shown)


aawsome:
I'm merging this and leave the open points to further PRs

PR #280, opened 2022-11-03
backup: Add --as-path option

aawsome:
The option `--as-path` lets rustic set an other path instead of the path specified by SOURCE. It also generates the tree structure according to this given path.

Example:
If you have the files `/dir/a` and `/dir/subdir/b` and call `rustic backup /dir`, the snapshot will contain
```
/dir
/dir/a
/dir/subdir
/dir/subdir/b
```
If you call `rustic backup /dir --as-path /my/directory`, it will generate
```
/my
/my/directory
/my/directory/a
/my/directory/subdir
/my/directory/subdir/b
```

### DIFF
--- a/src/archiver/archiver_impl.rs
+++ b/src/archiver/archiver_impl.rs
@@ -104,7 +104,13 @@ impl<BE: DecryptWriteBackend, I: IndexedBackend> Archiver<BE, I> {
         self.summary.total_dirsize_processed += size;
     }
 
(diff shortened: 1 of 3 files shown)


aawsome:
Added config file support, i.e. if you use `--as-path=/my-path` and provide a `[[backup.sources]]` section with `source = "/my-path"`, the options are taken from this section.

Issue #42, opened 2022-07-01
replace prettytable-rs?


aawsome:
As prettytable-rs doesn't seem to be actively maintained, see https://github.com/phsym/prettytable-rs/issues/139

This is related to #30.

JMarkin:
I think, good crates.
- [comfy-table](https://crates.io/crates/comfy-table)
- [tabled](https://crates.io/crates/tabled)

Checklist:
 - `should be free to use, of course (i.e. right licence)` all MIT
 - `being well maintained` < 1 month of last update
 - `having few other dependencies` from 2 to 3 required dependencies

I think comfy-table more similar to current code then tabled

What do you think about these crates ?

aawsome:
@JMarkin  When I had a quick lock, I also found those. So, yes, they are both good candidates. If you are willing, choose the one you think fits better (or the one you like more) and propose a PR. Thank you very much in advance!

aawsome:
closed by #290

PR #285, opened 2022-11-06
[WIP]: replace prettytable

JMarkin:
Hello, I tried to replace prettytable-rs with console. #42 
I tested it on repoinfo. Today, I changed only repoinfo. Result is on screen.

image

I am newable in Rust, but want to learn, please review)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1678,6 +1678,7 @@ dependencies = [
  "chrono",
  "clap",
  "clap_complete",
+ "console",
  "crossbeam-channel",
  "derivative",
  "derive-getters",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,6 +87,7 @@ humantime = "2"
 users = "0.11"
 itertools = "0.10"
 simplelog = "0.12"
+console = "0.15.2"
 
 [dev-dependencies]
 rstest = "0.15"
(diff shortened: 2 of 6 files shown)


aawsome:
Hey @JMarkin! Thanks a lot for tackling this issue!

So, for someone new in Rust, you did a very good job and the results look good.

There is just one thing - and I think I have to apologize for not being clear about my intent in #42.
When talking about maybe replacing prettytable I was thinking about checking if this crate really was a not-so-ideal choice and comparig with other crates who provide the same or superior functionality.

So the idea was to maybe exchange prettytable by a better suited crate.
In fact, I don't think that we should re-implement a table-display functionality in rustic given that there are many good crates out there. This is similar to the progress bars where I fully rely on indicatif which provides all the needed features or evolves to provide them.

For the ideal crate I think the following general conditions are important:
- should be free to use, of course (i.e. right licence)
- being well maintained
- having few other dependencies

Do you think you can also support rustic in that way and try out alternatives to prettytable?

JMarkin:
Sorry, I really didn't understand..
I read related issue and thought you approved console, I saw current tables and thought it is pretty simple for implement, now I understand mistakes my logic, thank you)

if need close pr, I will close.
 
Any way I want to complete with issue)

PR #290, opened 2022-11-11
issue-42: change prettytable -> comfy-table

JMarkin:
Hello, I started to move from prettytable to comfy-table. I set styles like previous. #42

image

Can I change other tables or you have some comments?


src/tui/table.rs
```diff
@@ -0,0 +1,53 @@
+use std::fmt::Display;
+
+use comfy_table::{presets::ASCII_MARKDOWN, Cell, ContentArrangement, Table};
+
+pub struct RusticTable {
+    table: Table,
```
aawsome, line 6: Do you know about the newtype pattern? i.e. you could easily use 
`pub struct RusticTable(Table)` and access table with `self.0`.

Or, maybe even better to just use an `type` definition and use the existing methods from `Table`?

src/tui/table.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub fn new() -> Self {
+        let mut table = Table::new();
+        table
+            .load_preset(ASCII_MARKDOWN)
+            .set_content_arrangement(ContentArrangement::Dynamic);
+
+        Self { table }
+    }
+
+    pub fn set_titles<I, D>(&mut self, titles: I)
```
aawsome, line 19: I would use the existing `set_header` from Table.
Moreover, this hard-encodes the format of the title - bold might be fine, but left/right-aligned surely doesn't fit for all tables. Better use the provided format methods at the place where the table is defined (i.e. in `repoinfo.rs`).

aawsome (changes requested):
Thanks @JMarkin for the start! comfytable looks good and I just had a quick view about your changes. Sorry for getting back so late - I've got a bunch of childs to look after ATM (including a new-born :baby: )

Here are my points:
- I would prefer using comfytable more directly. IMO there is no need to do a lot of abstraction (like own type).
- Besides that, sure, go on. Looking forward to seeing a finalized PR!

src/tui/table.rs
JMarkin: I removed abstraction table

JMarkin:
I completed, but found mystery. In snapshots or forget command, I see empty last line. I think it is bug on comfy table because `table.rows_iter().last()` doesn't show blank line... If you understand why, please tell me)

aawsome:
@JMarkin Thanks a lot, this looks really good already. 

I was a bit irritated that this adds a lot of dependencies, but these came all with the dependency crossterm. I think this should be ok.
A big plus is that the table render much nicer on small displays!

I did some changes and finetuning - see the last commit I added. Please tell me if you are fine with these changes and don't hesitate to ask if somethings feels unfamiliar or odd to you.

About the added lines: This was due to newlines in the paths/tags and also in the reasons given by `forget`. These seemed to be ignored by prettytable but are (correctly) displayed by comfy-table. So I consider this to be a bug in prettytable. I also corrected the newlines in my commit.

If you are fine with the adaptions, please rebase to current master and then we'll be able to merge this!

JMarkin:
@aawsome I merged current main branch into this and fixed conflict. I think need resolve conversation and merge.

aawsome:
I can look into how to rebase and merge. Did you habe a look at the changes I made? If you agree with them, I'll go ahead!

aawsome:
I rebased to main and squashed the commits into one.
@JMarkin if you rebase to main instead of merge main into your branch, you'll see exactly how main will look like after merging. This is why I prefer it that way.

PR #335, opened 2022-12-12
restore: Overwork treatment of existing files

### LINKED ISSUE #105: restore with existing files/dirs but wrong type or permission
Restore creates files/dirs or special files like symlinks during the restore process.

The standard case of an existing file also works: Its content is changed to the content-to-restore.

However, if files or dirs or special files (e.g. symlinks) exist but file type is wrong (e.g. existing dir, but we want to restore a file) or permissions do not allow to write to a file/dir, then errors or unexpected behavior happens.

There are two kind of error:
- errors which stop the resuming (most are errors in trying ro write to a file or to create a file)
- undesired behavior without getting an error. An example is an existing symlink where rustic seems to write to the symlink destination instead of removing the symlink an creating a new file.

### LINKED ISSUE #326: restore: Make reading of destination optional
Currently, `restore` reads the files in the destination dir to check if the contents are correct.
This could be skipped if we rely on metadata (like ctime/mtime/inode similar to the change detection in `backup` when backing up with a parent snapshot) or at lest be made optional.

See also https://github.com/restic/restic/pull/3425

### LINKED ISSUE #332: [ERROR] non-existing file, but restore algo returned existing file??!??
When doing a restore I get the following printed out

```
[ERROR] non-existing file, but restore algo returned existing file??!??
```

rustic v0.4.0

aawsome:
A rework of the treatment of existing files. 
The default is now to not read files with matching size and length (closes #326), but the new option `--ignore-mtime` is available to skip this check.

closes #332 
closes #105

### DIFF
--- a/src/backend/local.rs
+++ b/src/backend/local.rs
@@ -1,4 +1,4 @@
-use std::fs::{self, File};
+use std::fs::{self, File, OpenOptions};
 use std::io::{Read, Seek, SeekFrom, Write};
 use std::os::unix::fs::{symlink, FileExt, PermissionsExt};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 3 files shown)


aawsome:
@cameronbraid Thanks for testing! I also ran some checks, did a code review and all looks good.

PR #382, opened 2023-01-13
Avoid allocation for Id::{to,from}_hex

Dr-Emann:
I expect we use this pretty often, might be worth trying to optimize a bit.


aawsome (commented):
Good catch, @Dr-Emann 
I just tried to simplify `id.rs` a bit..

aawsome (approved):
LGTM!

PR #383, opened 2023-01-13
fix password-command

aawsome:
see #376 
This PR corrects the tokenization of the command string and adds debug output for the result.


aawsome:
The parser implementation might not be the best (first usage of the `nom` crate), but IMO it suffices to solve the password-command problem.
If anyone has a better solution, please propose a PR!

PR #386, opened 2023-01-14
Fix some clippy pedantic findings

Dr-Emann:
See individual commits for the specific clippy findings fixed


src/commands/prune.rs
```diff
@@ -589,9 +589,9 @@ impl Pruner {
         };
 
         self.repack_candidates.sort_unstable_by_key(|rc| rc.0);
-        let mut resize_packs: BlobTypeMap<Vec<_>> = Default::default();
-        let mut do_repack: BlobTypeMap<bool> = Default::default();
-        let mut repack_size: BlobTypeMap<u64> = Default::default();
+        let mut resize_packs: BlobTypeMap<Vec<_>> = BlobTypeMap::default();
+        let mut do_repack: BlobTypeMap<bool> = BlobTypeMap::default();
```
aawsome, line 593: Can't we remove the type hint here?

src/commands/prune.rs
```diff
@@ -589,9 +589,9 @@ impl Pruner {
...
+        let mut repack_size: BlobTypeMap<u64> = BlobTypeMap::default();
```
aawsome, line 594: Can't we remove the type hint here?

src/commands/prune.rs
```diff
@@ -589,9 +589,9 @@ impl Pruner {
         };
 
         self.repack_candidates.sort_unstable_by_key(|rc| rc.0);
-        let mut resize_packs: BlobTypeMap<Vec<_>> = Default::default();
-        let mut do_repack: BlobTypeMap<bool> = Default::default();
-        let mut repack_size: BlobTypeMap<u64> = Default::default();
+        let mut resize_packs: BlobTypeMap<Vec<_>> = BlobTypeMap::default();
```
aawsome, line 592: Is there a possibility to simply the type hint now that we are using `BlobTypeMap::default`?

aawsome (commented):
Thanks a lot @Dr-Emann for this PR.
Besides minor comments one question: Should we add some of these lints to `src/main.rs`

src/commands/prune.rs
Dr-Emann: Unfortunately, only `do_repack` can automatically be inferred as BlobTypeMap<bool>, leaving off the type hints for `resize_packs` and `repack_size` leads to a compile error that rust can't infer the type.

1. Would you prefer leaving off the type hint for do_repack?
2. We could use e.g. `BlobTypeMap::<Vec<_>>::default()`, which do you prefer?

aawsome: I removed all type hints, using your second option where needed

aawsome:
I updated the `default()` suggestions and reformated `main.rs` to make the format linter happy ;-)

aawsome (approved):
LGTM - thanks a lot @Dr-Emann, good work!

PR #395, opened 2023-01-17
Allow treating stdin as a tar archive

Dr-Emann:
This is a work in progress, but I'm planning to try to use this to stream a google takeout (which outputs multiple tar.gz archives, so I plan to concatenate them, un-gzip them, and pipe to rustic). This will allow me to backup my entire google takeout to backblaze without needing to store anything locally.

Similar to restic/restic#2226

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+
+[[package]]
 name = "adler"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -52,6 +58,12 @@ checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
 dependencies = [
  "memchr",
 ]
+
+[[package]]
+name = "aliasable"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "250f629c0161ad8107cf89319e990051fae62832fd343083bea452d93e2205fd"
 
 [[package]]
 name = "android_system_properties"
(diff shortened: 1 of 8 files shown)


aawsome:
Hi @Dr-Emann 
Just had a short look at this PR, but I'd say the general direction is perfectly right!

I also already thought about adding Traits for reading sources, where sources should provide a method to list files and a Reader to actually read the file content.

This could be implemented for a tar archive as well as for e.g. remote sources like an rclone source.
I think we should take the time within this PR to properly design and implement these traits for local sources and then for tar sources.

Dr-Emann:
Have to go back to the drawing board a little. It seemed to work, at first, but I noticed something in the debug logging output that really sucks:

Tar is just an unordered collection of entries, so you can get entries like:

```
file takeout/folder_a/file7
file takeout/folder_b/other_file
dir  takeout/folder_a/
file takeout/folder_a/file_1
```

Which causes the directory creation to be way off in rustic

aawsome:
Ok, this is a problem - the archiver expects directory entries to come ordered.

I already thought about changing the archiver code a bit:
- give the listed files (and the readers to read the contents) to a file archiver which will store the contents and return a tree entry.
- process the tree entries in a tree archiver

I had parallelization in mind (the file archiver needs no order and therefore file contents of different files can be processed in parallel). This is of course only true if you can also access the file readers in parallel (which most likely is not the case for tar archives).

But once the above is changed, we could use it for tar archives by processing the files (sequentially) and `collect()` the dir entries. Then sort the collected dir entries before processing them with the tree archiver.

I think it might be good to have the following properties in the Source trait:
- Can file reading be parallelized (true for local source, false for stdin and tar source)
- Do the dir entries come ordered (true for local source, n/a for stdin and false for tar souce)

aawsome:
@Dr-Emann Can you have a look at #425?

I copied the traits you are using here and reworked how the archiver is working into an Iterator approach.

I think if you start at top of 425, you can simply add to the `ReadSource` trait something like `const IN_ORDER: bool` and if this is set to `false` first collect all items after archiving the files, then sort by path and finally save the items in trees. 

The drawback would be that instead of sequentially processing the trees and saving them when finished, all items need to be saved in a `Vec` which could potentially increase memory usage of rustic. But if the items come out-of-order, this is the only way to solve it.

Actually I'm not completely finished with #425 and would highly appreciate a foreign review.

PR #403, opened 2023-01-23
Enable building on OpenBSD

bket:
This is needed to get rustic to build on OpenBSD. Lightly run tested the build against a local restic repo.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -47,6 +47,12 @@ jobs:
             architecture: i686
             binary-postfix: ""
             use-cross: true
+          - os: openbsd-latest
+            os-name: openbsd
+            target: x86_64-unknown-openbsd
+            architecture: x86_64
+            binary-postfix: ""
+            use-cross: true
 
     steps:
       - name: Checkout repository
--- a/src/backend/local.rs
+++ b/src/backend/local.rs
@@ -231,16 +231,16 @@ impl LocalBackend {
                 symlink(linktarget, filename)?;
             }
             NodeType::Dev { device } => {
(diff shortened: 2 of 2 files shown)


klemensn (approved):

aawsome:
Hi @bket 
Thank you very much for the PR! Would you mind adding OpenBSD targets in the CI/CD pipeline? 
Then builds are tested for OpenBSD and binaries are automatically released
`.github/workflows/release.yaml` is the right  place, I think we should add `x86_64-unknown-freebsd` or `x86_64-unknown-netbsd` as target - which is the right one?

bket:
> I think we should add `x86_64-unknown-freebsd` or `x86_64-unknown-netbsd` as target - which is the right one?

If I understand [https://doc.rust-lang.org/rustc/platform-support/openbsd.html](https://doc.rust-lang.org/rustc/platform-support/openbsd.html) correctly the target name should read $ARCH-unknown-openbsd, where $ARCH specifies the target processor architecture. I am not familiar with setting up a pipeline, so chances are that I missed something. For now I only added `x86_64-unknown-openbsd` as a target.

aawsome:
Ok, just realized that the release build pipeline doesn't run for "external" PRs. I'll just give it a try ;-)

aawsome (approved):
LGTM! I'm also not sure if the CI runs, but let's try it out!

aawsome:
Seems the toolchain support is missing to successfully.
I tried openbsd, netbsd and freebsd. #407 should add compile support for freebsd but binary builds were only possible for netbsd. For openbsd, there seems issues like https://github.com/rust-lang/rustup/issues/2168 which prevent the cross-compiling.

PR #438, opened 2023-02-20
WIP: Add docu

aawsome:


doc/020_installation.rst
```diff
@@ -0,0 +1,76 @@
...
+Unstable Builds
+===============
+
+Another option is to use the latest builds for the master branch, available on
+the `rustic beta download site
+<https://github.com/rustic-rs/rustic-beta>`__. These too are pre-compiled
+and ready to run, and a new version is built every time a push is made to the
+maina branch.
+
+From Source
```
LeoniePhiline, line 55: Can `cargo install` new used? This would be a good solution, requires no root perms and can be kept updated with `cargo-update`.

aawsome: `cargo install` should work. Don't know about `cargo-update` as I didn't use it so far. 
But also note that there is the `rustic self-update` command which downloads pre-compiled binaries from github.

LeoniePhiline: Yes `cargo install rustic-rs` worked (`cargo install rustic` failed because that's a different project 😩).

A self-update function is common on inferior platforms like windows, but on OSes where you have actual package management, using any kind of package manager is superior to self updating each app individually. The latter just does not scale.

E.g. I regularly run this:
```zsh
#!/usr/bin/zsh

set -eux

export PATH="/home/leonie/.cargo/bin:$PATH"

rustup self update
rustup update
cargo install-update --all
cargo cache --autoclean
```

And it updates all my many rust binaries:

```
atuin
cargo-audit
cargo-bloat
cargo-cache
cargo-changelog
cargo-deny
cargo-expand
cargo-generate
cargo-get
cargo-install-update
cargo-install-update-config
cargo-miri
cargo-outdated
cargo-public-api
cargo-release
cargo-sort
cargo-sweep
cargo-tarpaulin
cargo-watch
clippy-driver
delta
dua
grcov
hx
jql
rls
rust-gdb
rust-gdbgui
rust-lldb
rustic
rustup
sk
```

When rustic is packaged by my linux distribution, then I will switch to their well-maintained packages. Until then I must rely on the `cargo-install-update` cargo command supplied by the [`cargo-update`](https://crates.io/crates/cargo-update) crate.

IMHO, if building from _git_ source is documented (which only makes sense if you want to run the latest `main` branch), then building _released versions_ from source and keeping up to date simply via `cargo install` should be documented too. :)

doc/030_preparing_a_new_repo.rst
```diff
@@ -0,0 +1,250 @@
...
+
+   Remembering your password is important! If you lose it, you won't be
+   able to access data stored in the repository.
+
+.. warning::
+
+   On Linux, storing the backup repository on a CIFS (SMB) share is not
+   recommended due to compatibility issues. Either use another backend
+   or set the environment variable `GODEBUG` to `asyncpreemptoff=1`.
+   Refer to GitHub issue `#2659 <https://github.com/rustic/rustic/issues/2659>`_ for further explanations.
```
istudyatuni, line 86: It's possible that this is not relevant to rustic, but I not checked. Also, correct url is https://github.com/restic/restic/issues/2659

PR #532, opened 2023-03-25
OpenBSD does not support extended file attributes

bket:
#522 breaks building on OpenBSD because there is no xattr support.


aawsome:
Thanks a lot for your PR! Unfortunately I didn't manage to build OpenBSD in the CI.
If you see a way to add it, please open a PR!

I fixed and reformatted this PR. If the CI compiles it, I'll merge it.

PR #603, opened 2023-04-23
forget: Interpret '--keep-X -1' as 'keep all X'

thndrbrrr:
This PR interprets the use of value `-1` (negative 1) for `--keep-*` options to mean "keep all".

So `--keep-hourly -1` would mean "keep all hourly snapshots" etc:

```shell
$ ./target/release/rustic -r ~/rstc1/ forget --prune -n --keep-hourly -1
using no config file (./rustic.toml doesn't exist)
enter repository password: 
[INFO] repository local:/Users/thndrbrrr/rstc1/: password is correct.
[INFO] using cache at /Users/thndrbrrr/Library/Caches/rustic/a35ef0a3a66303a26a652040f5eb53d471fac3cf33779126b960c33184b59621
snapshots for (host [EP-ML-10471], label [], paths [/Users/thndrbrrr/devel/misc/thndrbrrr/rustic/test-bats])

| ID       | Time                | Host        | Label | Tags | Paths                                                              | Action | Reason |
|----------|---------------------|-------------|-------|------|--------------------------------------------------------------------|--------|--------|
| 564a8fc1 | 2023-04-18 15:11:01 | EP-ML-10471 |       |      | /Users/thndrbrrr/devel/misc/thndrbrrr/rustic/test-bats | keep   | hourly |
| cb51365f | 2023-04-18 14:55:39 | EP-ML-10471 |       |      | /Users/thndrbrrr/devel/misc/thndrbrrr/rustic/test-bats | keep   | hourly |
| 6556f929 | 2023-04-18 14:40:39 | EP-ML-10471 |       |      | /Users/thndrbrrr/devel/misc/thndrbrrr/rustic/test-bats | remove |        |
| b295de85 | 2023-04-18 14:25:38 | EP-ML-10471 |       |      | /Users/thndrbrrr/devel/misc/thndrbrrr/rustic/test-bats | remove |        |
| 71757a83 | 2023-04-18 14:10:37 | EP-ML-10471 |       |      | /Users/thndrbrrr/devel/misc/thndrbrrr/rustic/test-bats | remove |        |
| 4a0f8eda | 2023-04-18 13:55:37 | EP-ML-10471 |       |      | /Users/thndrbrrr/devel/misc/thndrbrrr/rustic/test-bats | keep   | hourly |
| 682089ba | 2023-04-18 13:40:36 | EP-ML-10471 |       |      | /Users/thndrbrrr/devel/misc/thndrbrrr/rustic/test-bats | remove |        |
| 2aba13bd | 2023-04-18 13:25:36 | EP-ML-10471 |       |      | /Users/thndrbrrr/devel/misc/thndrbrrr/rustic/test-bats | remove |        |
| 395d7796 | 2023-04-18 13:10:35 | EP-ML-10471 |       |      | /Users/thndrbrrr/devel/misc/thndrbrrr/rustic/test-bats | remove |        |
| 391f52e2 | 2023-04-18 13:09:23 | EP-ML-10471 |       |      | /Users/thndrbrrr/devel/misc/thndrbrrr/rustic/test-bats | remove |        |
| 511e4a5e | 2023-04-17 15:03:00 | EP-ML-10471 |       |      | /Users/thndrbrrr/devel/misc/thndrbrrr/rustic/test-bats | keep   | hourly |

would have removed the following snapshots:
 [6556f929ecd9ee122c60dd3eb3d4e972ed9481ffca2ef25607789d8859e6f4b9, b295de85959ef6778a988b0156f9dd5611faa3e05f2350a27f507d126dbede01, 71757a83f6a3fcdc52614b39440dbbf021149eb88a4bedb776dab19efa7810b1, 682089baca3cff22fdc2ee3f37041fa227e793b08b1e0e095d252d349479f8f9, 2aba13bd1116106e64b57bb95f2a5cbb9e43c52548cb6f82aba4bffafe7a6204, 395d779637861794ff6e2facaa863913ece2ce5827a4d5820b67c2830232913e, 391f52e29469a8d2724417619c910b2ec8223ea92cd57212d21690347b717513]
[00:00:00] reading index...               ████████████████████████████████████████          1/1                                                                                                  
[00:00:00] reading snapshots...           ████████████████████████████████████████          5/5                                                                                                  
[00:00:00] finding used blobs...          ████████████████████████████████████████          5/5                                                                                                  
[00:00:00] getting packs from repository...                                                                                                                                                      to repack:          0 packs,          0 blobs,        0 B
this removes:                         0 blobs,        0 B
to delete:          0 packs,          0 blobs,        0 B
total prune:                          0 blobs,        0 B
remaining:                          486 blobs,   14.5 MiB
unused size after prune:        0 B (0.00% of remaining size)

packs marked for deletion:          0,        0 B
 - complete deletion:               0,        0 B
 - keep marked:                     0,        0 B
 - recover:                         0,        0 B
```

The usage instructions have been updated accordingly:

```shell
$ ./target/release/rustic forget --help
rustic-forget 
Remove snapshots from the repository

USAGE:
    rustic forget [OPTIONS] [IDS]...

[---8<---snip---8<---]

RETENTION OPTIONS:
        --keep-tags <TAG[,TAG,..]>
            Keep snapshots with this taglist (can be specified multiple times)

        --keep-id <ID>
            Keep snapshots ids that start with ID (can be specified multiple times)

    -l, --keep-last <N>
            Keep the last N snapshots (N == -1: keep all snapshots) [default: 0]

    -H, --keep-hourly <N>
            Keep the last N hourly snapshots (N == -1: keep all hourly snapshots) [default: 0]

    -d, --keep-daily <N>
            Keep the last N daily snapshots (N == -1: keep all daily snapshots) [default: 0]

[---8<---snip---8<---]
```

### DIFF
--- a/changelog/new.txt
+++ b/changelog/new.txt
@@ -9,6 +9,7 @@ Bugs fixed:
 
 New features:
 - backup: Backing up (small) files is now much more parallelized.
+- forget: Using "-1" as value for --keep-* options will keep all snapshots of that interval
 - prune: Added option --repack-all
 - Option --dry-run is now a global option and can also be defined in the config file or via env variable 
 - Updated to clap v4
--- a/src/commands/forget.rs
+++ b/src/commands/forget.rs
@@ -175,45 +175,45 @@ pub(super) struct KeepOptions {
     #[merge(strategy=merge::vec::overwrite_empty)]
     keep_ids: Vec<String>,
 
-    /// Keep the last N snapshots
-    #[clap(long, short = 'l', value_name = "N", default_value = "0")]
+    /// Keep the last N snapshots (N == -1: keep all snapshots)
+    #[clap(long, short = 'l', value_name = "N", default_value = "0", allow_hyphen_values = true, value_parser = clap::value_parser!(i32).range(-1..))]
     #[merge(strategy=merge::num::overwrite_zero)]
-    keep_last: u32,
+    keep_last: i32,
 
-    /// Keep the last N hourly snapshots
-    #[clap(long, short = 'H', value_name = "N", default_value = "0")]
+    /// Keep the last N hourly snapshots (N == -1: keep all hourly snapshots)
+    #[clap(long, short = 'H', value_name = "N", default_value = "0", allow_hyphen_values = true, value_parser = clap::value_parser!(i32).range(-1..))]
     #[merge(strategy=merge::num::overwrite_zero)]
-    keep_hourly: u32,
+    keep_hourly: i32,
 
-    /// Keep the last N daily snapshots
-    #[clap(long, short = 'd', value_name = "N", default_value = "0")]
+    /// Keep the last N daily snapshots (N == -1: keep all daily snapshots)
+    #[clap(long, short = 'd', value_name = "N", default_value = "0", allow_hyphen_values = true, value_parser = clap::value_parser!(i32).range(-1..))]
     #[merge(strategy=merge::num::overwrite_zero)]
-    keep_daily: u32,
+    keep_daily: i32,
 
-    /// Keep the last N weekly snapshots
-    #[clap(long, short = 'w', value_name = "N", default_value = "0")]
+    /// Keep the last N weekly snapshots (N == -1: keep all weekly snapshots)
+    #[clap(long, short = 'w', value_name = "N", default_value = "0", allow_hyphen_values = true, value_parser = clap::value_parser!(i32).range(-1..))]
     #[merge(strategy=merge::num::overwrite_zero)]
-    keep_weekly: u32,
+    keep_weekly: i32,
 
-    /// Keep the last N monthly snapshots
-    #[clap(long, short = 'm', value_name = "N", default_value = "0")]
+    /// Keep the last N monthly snapshots (N == -1: keep all monthly snapshots)
+    #[clap(long, short = 'm', value_name = "N", default_value = "0", allow_hyphen_values = true, value_parser = clap::value_parser!(i32).range(-1..))]
     #[merge(strategy=merge::num::overwrite_zero)]
-    keep_monthly: u32,
+    keep_monthly: i32,
 
-    /// Keep the last N quarter-yearly snapshots
-    #[clap(long, value_name = "N", default_value = "0")]
+    /// Keep the last N quarter-yearly snapshots (N == -1: keep all quarter-yearly snapshots)
+    #[clap(long, value_name = "N", default_value = "0", allow_hyphen_values = true, value_parser = clap::value_parser!(i32).range(-1..))]
     #[merge(strategy=merge::num::overwrite_zero)]
-    keep_quarter_yearly: u32,
+    keep_quarter_yearly: i32,
 
-    /// Keep the last N half-yearly snapshots
-    #[clap(long, value_name = "N", default_value = "0")]
+    /// Keep the last N half-yearly snapshots (N == -1: keep all half-yearly snapshots)
+    #[clap(long, value_name = "N", default_value = "0", allow_hyphen_values = true, value_parser = clap::value_parser!(i32).range(-1..))]
     #[merge(strategy=merge::num::overwrite_zero)]
-    keep_half_yearly: u32,
+    keep_half_yearly: i32,
 
-    /// Keep the last N yearly snapshots
-    #[clap(long, short = 'y', value_name = "N", default_value = "0")]
+    /// Keep the last N yearly snapshots (N == -1: keep all yearly snapshots)
+    #[clap(long, short = 'y', value_name = "N", default_value = "0", allow_hyphen_values = true, value_parser = clap::value_parser!(i32).range(-1..))]
     #[merge(strategy=merge::num::overwrite_zero)]
-    keep_yearly: u32,
+    keep_yearly: i32,
 
     /// Keep snapshots newer than DURATION relative to latest snapshot
     #[clap(long, value_name = "DURATION", default_value = "0h")]
@@ -404,10 +404,12 @@ impl KeepOptions {
 
         for (check_fun, counter, reason1, within, reason2) in keep_checks {
             if !has_next || last.is_none() || !check_fun(sn, last.unwrap()) {
-                if *counter > 0 {
-                    *counter -= 1;
+                if *counter != 0 {
                     keep = true;
                     reason.push(reason1);
+                    if *counter > 0 {
+                        *counter -= 1;
+                    }
                 }
                 if sn.time + Duration::from_std(*within).unwrap() > latest_time {
                     keep = true;


aawsome:
@thndrbrrr Thanks a lot for your PR - nice idea to add a "keep all" possibility!

I have two remarks/wishes:
- I don't know if `-1` is the best user experience to specify "keep all". I would like something like `--keep-daily all` much better. This could be e.g. implemented using a newtype and implementing `FromStr` for it. But I'm not sure about the config file - here we then should allow `keep-daily = 5` as well as `keep-daily = "all"`, and I don' t really know how this would be accomplished when parsing the TOML . What do you think about this idea?
- Please add a line to the file `changelog/new.txt`. Thanks!

thndrbrrr:
@aawsome Hahaha 😆 I initially had the same thought ... `all` would be clearer and not so easy to confuse with another option. What I like about the current implementation though is that the changes are really minimal.

At the same time I'm thinking it's probably worthwhile to look into supporting `all` as a value, like you said maybe through a struct with a `From<u16>` and `From<String>` implementation. Just gotta figure out the TOML parsing. And since it would be a custom struct then having a function like `fn checked_decrement(&mut self) -> bool` to improve the code that uses it might also be nice:

```rust
if keep_counter.checked_decrement() {
    keep = true;
    reason.push(reason1);
}
```
`checked_decrement()` would always work for `all`, and only return `false` if the counter was at `0`. Something like that.

Here's a thought: the current implementation could actually be left as is, and the future implementation with `all` would still support `-1` for backwards compatibility. Lemme know what you think. :)

aawsome:
> Here's a thought: the current implementation could actually be left as is, and the future implementation with `all` would still support `-1` for backwards compatibility.

Agree :+1: 
Can you please add a line in `new.txt` (and squash into one commit) and then let us merge this!

thndrbrrr:
Done and done!

aawsome (approved):
LGTM - Thanks again!

aawsome (approved):
clippy claimed a missing `;` - which I just added.

aawsome:
About adding the `all` functionality using a custom type: I think for Clap a `FromStr` impl would suffice. For Serde/TOML I think `Deserialize` must be implemented using a suitable `Visitor`.

thndrbrrr:
I was able to get the CLI parsing to work with something like:

```rust
pub(super) struct KeepOptions {
    #[clap(long, short = 'H', value_name = "N", default_value = "0", allow_hyphen_values = true, value_parser = parse_counter_i32)]
    #[merge(strategy=merge::num::overwrite_zero)]
    keep_hourly: i32,

    # ...
}

fn parse_counter_i32(val: &str) -> Result<i32, &'static str> {
    match val.parse::<i32>() {
        Ok(n) if n >= -1 => Ok(n),
        _ => match val.trim().to_lowercase().as_str() {
            "all" => Ok(-1),
            _ => Err("Value must be >= -1 or 'all'."),
        },
    }
}
```

... but that doesn't solve TOML parsing. I looked into that as well, testing things out with of a defined `struct`, and it works, but only when the value is a string. So `keep_hourly = "all"` and `keep_hourly = "5"` work, but `keep_hourly = 5` doesn't, which is unintuitive. There's a variation that uses an `enum` to determine the type, also with a `Deserializer`, and that one works, but I don't know how this would fit with the `RusticConfig` implementation.

Unfortunately I wasn't able to figure out how to get a `Visitor` working with `RustConfig` parsing.

aawsome:
Actually, I would make a newtype like  `struct CountOrAll(i32)` and implement `FromStr` for this type (like you did with your parse function)
Then it would be sufficent for clap to have `keep_hourly : CountOrAll`.

Moreover, for this newtype I think `Deserialize` must be manually implemented. For how to manually implement it, have a look at https://serde.rs/impl-deserialize.html - I think this would involve implementing `Visitor` for a self-defined `CountOrAllVisitor` in a way that the `visit_i32` as well as the `visit_str` methods are implemented. But this is just a guess - in fact it should depend on which `visit_*` the toml crate calls...

thndrbrrr:
I think I got it, let me know if this looks right:

```rust
#[derive(Debug, Clone, PartialEq)]
struct CountOrAll(i32);

impl FromStr for CountOrAll {
    type Err = String;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s.parse::<i32>() {
            Ok(n) if n >= -1 => Ok(Self(n)),
            _ => match s.trim().to_lowercase().as_str() {
                "all" => Ok(Self(-1)),
                _ => Err(String::from("value must be >= -1 (i32) or 'all' (str)")),
            },
        }
    }
}

impl<'de> Deserialize<'de> for CountOrAll {
    fn deserialize<D>(deserializer: D) -> Result<CountOrAll, D::Error>
        where D: serde::de::Deserializer<'de>
    {
        struct CountOrAllVisitor;

        impl<'de> serde::de::Visitor<'de> for CountOrAllVisitor {
            type Value = CountOrAll;

            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
                formatter.write_str("an integer >= -1 or string 'all' (case insensitive)")
            }

            fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
                where
                    E: serde::de::Error, {
                match CountOrAll::from_str(s) {
                    Ok(n) => Ok(n),
                    Err(msg) => Err(serde::de::Error::custom(msg))
                }
            }

            fn visit_i64<E>(self, v: i64) -> Result<Self::Value, E>
                where
                    E: serde::de::Error, {
                        self.visit_str(&v.to_string())
            }
        }

        deserializer.deserialize_str(CountOrAllVisitor)
    }
}

#[derive(Debug, Parser, Deserialize)]
struct Cli {
    #[arg(short, long, default_value = "0", allow_negative_numbers = true)]
    count: CountOrAll,
}

fn main() {
    assert_eq!(Cli::parse_from(vec!["program", "--count", "2"]).count, CountOrAll(2));
    assert_eq!(Cli::parse_from(vec!["program", "--count", "-1"]).count, CountOrAll(-1));
    assert_eq!(Cli::parse_from(vec!["program", "--count", r#"all"#]).count, CountOrAll(-1));
    assert_eq!(toml::from_str::<Cli>(r#"count = 2"#).unwrap().count, CountOrAll(2));
    assert_eq!(toml::from_str::<Cli>(r#"count = "2""#).unwrap().count, CountOrAll(2));
    assert_eq!(toml::from_str::<Cli>(r#"count = -1"#).unwrap().count, CountOrAll(-1));
    assert_eq!(toml::from_str::<Cli>(r#"count = "-1""#).unwrap().count, CountOrAll(-1));
    assert_eq!(toml::from_str::<Cli>(r#"count = "all""#).unwrap().count, CountOrAll(-1));
    assert_eq!(toml::from_str::<Cli>(r#"count = "ALL""#).unwrap().count, CountOrAll(-1));
}
```

aawsome:
Yeah, looks really good!

Looking at your `Deserialize` impl, I realized that is actually is quite generic and doesn't rely at all on `CountOrAll` specialities - I'm wondering if this (Deserialize a given string, if it is a number, give as string using `.to_string()`) isn't already implemented in some crate which we could use...

Anyway, you can propose a PR and we'll add this functionality!

aawsome:
I found https://docs.rs/serde-this-or-that/0.4.2/serde_this_or_that/fn.as_string.html
You could use this and do
```
impl<'de> Deserialize<'de> for CountOrAll {
    fn deserialize<D>(deserializer: D) -> Result<CountOrAll, D::Error>
        where D: serde::de::Deserializer<'de>
  {
    let s = as_string(deserializer)?;
    CountOrAll::from_str(s).map_err(serde::de::Error::custom)
  }
}

**Comment by thndrbrrr** [comment]
Thanks for those comments. Would look like this then:

```rust
#[derive(Default, Debug, Clone, PartialEq)]
enum CountOrAll{
    #[default]
    NotSet,
    All,
    Count(u32)
}

impl FromStr for CountOrAll {
    type Err = String;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s.parse::<i32>() {
            Ok(n) if n >= 0 => Ok(Self::Count(n.try_into().unwrap())),
            Ok(n) if n == -1 => Ok(Self::All),
            _ => match s.trim().to_lowercase().as_str() {
                "" => Ok(Self::NotSet),
                "all" => Ok(Self::All),
                _ => Err(String::from("value must be >= -1 (i32) or 'all' (str)")),
            },
        }
    }
}

impl<'de> Deserialize<'de> for CountOrAll {
    fn deserialize<D>(deserializer: D) -> Result<CountOrAll, D::Error>
        where D: serde::de::Deserializer<'de>
    {
        let s = serde_this_or_that::as_string(deserializer)?;
        CountOrAll::from_str(&s).map_err(serde::de::Error::custom)
    }
}

#[derive(Debug, Parser, Deserialize)]
struct Cli {
    #[arg(short, long, default_value = "", allow_negative_numbers = true)]
    count: CountOrAll,
}

fn main() {
    assert_eq!(CountOrAll::default(), CountOrAll::NotSet);
    assert_eq!(Cli::parse_from(vec!["program"]).count, CountOrAll::NotSet);
    assert_eq!(Cli::parse_from(vec!["program", "--count", "2"]).count, CountOrAll::Count(2));
    assert_eq!(Cli::parse_from(vec!["program", "--count", "-1"]).count, CountOrAll::All);
    assert_eq!(Cli::parse_from(vec!["program", "--count", r#"all"#]).count, CountOrAll::All);
    assert_eq!(toml::from_str::<Cli>(r#"count = 2"#).unwrap().count, CountOrAll::Count(2));
    assert_eq!(toml::from_str::<Cli>(r#"count = "2""#).unwrap().count, CountOrAll::Count(2));
    assert_eq!(toml::from_str::<Cli>(r#"count = -1"#).unwrap().count, CountOrAll::All);
    assert_eq!(toml::from_str::<Cli>(r#"count = "-1""#).unwrap().count, CountOrAll::All);
    assert_eq!(toml::from_str::<Cli>(r#"count = "all""#).unwrap().count, CountOrAll::All);
    assert_eq!(toml::from_str::<Cli>(r#"count = "ALL""#).unwrap().count, CountOrAll::All);
}
```

A couple of notes:

- Parsing still assumes the value is an `i32` in order to support `-1` for backwards compatibility ... but: we might just consider getting rid of this now. It's been just a few days, and there's no need for backwards compatibility for something that (presumably) hasn't been used so far.
- `clap` wants a default value, otherwise `count` becomes a required option. Right now the empty string `default_value = ""` is used for that, which will be parsed as `Count::NotSet`. Not entirely sure I like it, see alternative below.

The default value could also be specified with `default_value_t = CountOrAll::NotSet`, but that in turn requires implementing `fmt::Display`:

```rust
impl std::fmt::Display for CountOrAll {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        let d = match *self {
            Self::NotSet => String::from(""),
            Self::All => String::from("all"),
            Self::Count(n) => n.to_string()
        };
        write!(f, "{}", d)
    }
}
```

This seems again like lots of code for a trivial use case ... maybe the empty string being the default value as shown above is a simpler solution, and it works just as well. Lemme know what you think!

PR #617, opened 2023-04-26
Refactor into `rustic_core` library, `rustic` cli library and `rustic` application part

### LINKED ISSUE #461: Split rustic into library and binary crate

simonsan:
Fixes #461 

# Todo
## Library
- [x] replace removed `anyhow` result with suited `ErrorKinds`
  - [x] mostly done, but `rustic_core/backend/stdin.rs` is missing
- [x] go through ErrorKinds that contain `ErrorKindMember(#[from] AnotherError)` and check if we can handle them within code and use our own Error for it 
- [x] replace `anyhow` `bail!` macro at places with returning library errors
  - [x] research: which errors does the library part expose? (search for usage of `bail!` macro)
- [X] introduce feature flags for `clap`, `merge` -> combine them under `cli` feature flag
- [x] refactor/reimplement logging/tracing

### Public API
- [X] Added a integration test case for the public API: <https://github.com/Enselic/cargo-public-api#-as-a-ci-check>
  - run with: 
    - Linux: `export UPDATE_EXPECT=1; cargo test public_api -p rustic_core`
    - Windows (Powershell): `$env:UPDATE_EXPECT=1; cargo test public_api -p rustic_core`  
- <https://www.lurklurk.org/effective-rust/re-export.html>
- [X] <https://web.archive.org/web/20230128000646/http://www.tglman.com/posts/rust_lib_error_management.html>

## CLI
- [x] reimplement commands based on `Runnable` trait
  - [x] check `src/commands/start.rs` for implementation details
- [x] implement config file
- [x] implement merging strategies
- [x] check `helpers.rs` for issues to be tackled

## CI
- [x] Check workflows if and how they need adaptation
- [x] create workflow for `public api` check
  - utilizing existing `public_api` integration test for various OS

# Follow up
- progress bars: how to only in CLI not in library (trait `ProgressExt`)
  - maybe only implement them in CLI and have a `progress` trait in library, that certain structs can implement, so if something in that regards is needed, each struct can expose their own progress/status 
- check for non-consumed parameters (activate clippy lints)
- move outputs of e.g. `println!` and `eprintln!` to structs

## Public API
- document public API well
- <https://deterministic.space/elegant-apis-in-rust.html>
- <https://rust-lang.github.io/api-guidelines/>

## CI
- determine workflow for releasing packages
  - e.g. with <https://github.com/marcoIeni/release-plz>
  - or `cargo-smart-release`

# Research
empty

# Notes
- 2023-05-02: we discussed goals of this PR and limited the scope to:
  - splitting the library and cli part
  - implement the error handling needed
  - reimplement the commands on abscissa framework
  - check if logging is working
  - implement overriding/merging config internally
  - we have discussed as well, that the library won't be made public (as in published to crates.io for now), as it needs to be further refined after the split


aawsome:
Hi @simonsan 
Just had a short glance at abscissa but I think I didn't yet fully get how the framework exactly works.
What I think I understood:
- we should use library error handling like you already started to implement (Thanks a lot!)
- logging should be already working as abscissa supports the log crate
- the CLI command structure is already pretty good, everything already lives under `src/commands`
- abscissa uses a config, pretty similar to what rust uses in `src/commands/configfile.rs`and mainly some renaming/moving must be done.
- It is right that rustic currently uses the same structs for defining config file sections and CLI options and therefore derives `clap::Parser` as well as `Deserialize`. I think we should keep this in the first run. We can later decide if we should change something. The good point is that the commands which do config overwrites have all struct which also derive `merge::Merge` and therefore `config::Overwrite` can be easily implemented by just calling `merge` from the `Merge` trait.

Some points that may need some thinking:
- There are progress bars within many functions which is currently implemented by generating and passing an `indicatif::ProgressBar` struct. I'd think that in order to use in a library we should make this generic over some `Progress` trait which is trivially implemented by `indicatif::ProgressBar` and use `indicatif::ProgressBar` only in the CLI command(s).
- There are few other topics in the `commands/helper.rs` which need to be tackled.
- Besides progress bars and logs, some commands do produce output, e.g. `backup`, `snapshots`, `repoinfo`, `forget`, `prune`, see the `println!` usage. This should be moved to suitable structs (many are obvious like `backup` producing a `SnapshotFile`, `snapshots` producing a `Vec<SnapshotFile>`)

Thanks again for tackling this issue and don't hesitate to call out for help or directions!

aawsome:
> Any ideas what's better, because I think it's really important to decouple there. Keeping the `clap` part in a library does ~doesn't make my brain super happy~ gives me a bit of headache, tbf :D

My suggestion: 
The `Parser`/`Args` impl should be hidden behind a `clap` feature flag and `Deserialize` impl should be hidden behind a `serde` feature flag (and maybe the same for `Merge`/`merge`) for the config option structs which we expose in a library crate. The  CLI crate could then use the library with this features and doesn't have to redefine them.
Another crate just using the the library could manually set the option structs. I would also make all fields `pub` and impl `Default` for these structs, so it can be easily used. Another possibility, of course, would be to implement the builder pattern.

> EDIT: But also in general I wonder how the approach is for (De)-Serialization? Especially the `serde_as` seems important?

`serde_as` is currently mainly used when struct don't implement `Deserialize` - mostly when `FromStr` is already implemented - or for binary data that is in fact serialized as BASE64.

> Are you reachable via a platform that allows faster communication than this one, e.g. Discord, Matrix, IRC, Telegram?

I'll look for a way to communicate. However, I may be reachable only at some awkward times as I have some other responsibilities like job and family/children... ;-)

aawsome:
Ah, just saw one thing: The structs under `repofiles` should still always implement `Serialize` and `Deserialize` as the repository format is JSON and we need to read/write to the repository also when called as library.

aawsome:
So maybe it doesn't make sense to add the `serde` feature, as we will anyway depend on serde.

aawsome:
> Status update: Library builds now on openSuse Tumbleweed. rocket

Wow! Great!

About interface tests:

> This is the full output though, over 5k lines. Used for CI basically. 
> Could you please take a look at it and give an opinion what needs to be changed there?

Uhmm - actually didn't read through the 5k lines :-) I also don't know if we should expose really all this as public interface. Most likely not..

Generally about the interface:
In the commands there is currently loads of logic. I actually would start with only exposing this functionality to a public interface, i.e. something like `rustic_core::snapshots(..)`, `rustic_core::backup(..)` and so on...
This would then be a very tight library interface which could be extended later step-by-step with more lower-level APIs (like manually opening a repository or later defining a user-defined backend and use this to open a repository, ...)
In the first run a CLI would then only do very simple things like parsing CLI options, setting up the Progressbars and just call the exposed library functionality.
If you like I can sketch this in a separate PR.

Also, I'm wondering that now the error types have been overworked if it would make sense to just use this in a first-start PR which we could merge pretty fast. Thinking about getting to the split library/CLI step-by-step...

The next days I unfortunately have only few time - weekend is family-time and in Germany the 1st of May is a holiday...

aawsome:
I created a Discord channel: https://discord.gg/zGZsfkZue6
This evening I'll try to be there at around 21:15 MET (depends on when kids are in bed).

I never used Discord before, so hope everything works out ;-)

aawsome (approved):
Thanks a lot @simonsan for all your work!

I checked the changes today and didn't find problems - however as this is a big redesign, there might be still some bugs lurking around. We'll keep this beta for a while to find those bugs - if anyone finds something suspicious, please don't hesitate to report an issue!

LGTM!

PR #621, opened 2023-04-28
Fix typo

hirschenberger:
Fix typo


aawsome (approved):
Good catch! Thanks for the PR!

PR #624, opened 2023-04-30
restore: Download multiple contiguous blobs in one request

aawsome:
The restore command used to download each needed blob by its own.
This PR checks if there are many needed blobs which form a contiguous piece of data within a pack file and downloads them together.
This should reduce repository operations and speed-up the restore especially for remote/high latency backends.
Also the restore path is no longer saved once per blob in the file, but only once per file. So this PR should reduce memory for large files-to-restore.


aawsome:
rebased

PR #672, opened 2023-06-12
Use global config files when user-specific file doesn't exist

### LINKED ISSUE #652: Support global configuration files in /etc/rustic/
I think it would be great to have global config files in `/etc/rustic/` which could be used for backups by "system" users (which don't have a home directory) or configurations shared between users, so for example one user could create the backup and other users would still be able to access it.

The configuration files at `~/.config/` would have precedence over `/etc/`.

If you're interested, I'm willing to find some time and give this a try.

pheki:
Fixes #652 

For all profile names, it will look for configuration files in the following paths and in the following order:

- User-specific config path (same as it always was), e.g. `~/.config/rustic/{profile_name}.toml` on linux
- Global config path: `/etc/rustic/{profile_name}.toml` on Linux and MacOS, `%PROGRAMDATA%\config\rustic\{profile_name}.toml` on Windows
- Current directory: `./{profile_name}.toml`

It will just use the first found file, meaning that if you have a config in the user-specific config path, it will override the config with the same name in the global config path, ignoring everything on it. 

`use-profile` works normally, meaning you can have a user-specific config "inheriting" from the global config.

On smaller behavior changes, I removed the check if the directory exists, it will just check if the file exists directly, and I've fixed  `./{profile-name}.toml` (it was giving an error as it wasn't a canonical path, just had to change `AbsPathBuf::new` to `AbsPathBuf::canonicalize`). 

I haven't been able to test this on MacOS yet.

Is this how you think it should work?

# Example outputs

## Config files

`/etc/rustic/apple.toml`
```toml
[global]
```

`~/.config/rustic/banana.toml`
```toml
[global]
use-profile = ["apple"]
```

`./cocoa.toml`
```toml
[global]
use-profile = ["banana"]
```

## Outputs

```console
$ cargo run --quiet -- -P cocoa ls a
using config ./cocoa.toml
using config /home/bilko/.config/rustic/banana.toml
using config /home/bilko/.config/rustic/apple.toml
error: No repository given. Please use the --repository option.
```

```console
$ cargo run --quiet -- -P dragon-fruit ls a
using no config file, none of these exist: /home/bilko/.config/rustic/dragon-fruit.toml, /etc/rustic/dragon-fruit.toml, ./dragon-fruit.toml
error: No repository given. Please use the --repository option.
```

Let's change `banana.toml` to point to `apricot` instead:

```console
$ cargo run --quiet -- -P cocoa ls a
using config ./cocoa.toml
using config /home/bilko/.config/rustic/banana.toml
using no config file, none of these exist: /home/bilko/.config/rustic/apricot.toml, /etc/rustic/apricot.toml, ./apricot.toml
error: No repository given. Please use the --repository option.
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -6,7 +6,7 @@
 
 pub(crate) mod progress_options;
 
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 
 use directories::ProjectDirs;
 
@@ -16,6 +16,7 @@ use abscissa_core::config::Config;
 use abscissa_core::path::AbsPathBuf;
 use abscissa_core::FrameworkError;
 use clap::Parser;
+use itertools::Itertools;
 use rustic_core::RepositoryOptions;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -50,29 +50,33 @@ pub struct RusticConfig {
...
             eprintln!("using config {}", path.display());
-            let mut config = Self::load_toml_file(AbsPathBuf::new(&path)?)?;
+            let mut config = Self::load_toml_file(AbsPathBuf::canonicalize(path)?)?;
             // if "use_profile" is defined in config file, merge the referenced profiles first
             for profile in &config.global.use_profile.clone() {
                 config.merge_profile(profile)?;
             }
             self.merge(config);
         } else {
+            let paths_string = paths.iter().fold(String::new(), |acc, path| {
```
aawsome, line 67: Why not use `join` from itertools? (It is already in the dependency list)

pheki: Done. Nice catch! Didn't know it was already included

aawsome (approved):
Thanks a lot @pheki for your work!
I rebased and squashed this PR.

LGTM!

PR #698, opened 2023-06-22
feat: option to disable requiring git repository for git-ignore rules

istudyatuni:


aawsome (changes requested):
Thanks a lot!
Can you please add an entry in `changelog/new.txt` (and squash everything to 1 commit after)?

aawsome (approved):
LGTM

PR #699, opened 2023-06-22
feat: wait for password-command to exit

istudyatuni:
Useful for commands that requires interactive input

tested on Linux

### DIFF
--- a/changelog/new.txt
+++ b/changelog/new.txt
@@ -10,3 +10,4 @@ New features:
 - restore: The restore algorithm has been improved and should now be faster for remote repositories.
 - restore: Files are now allocated just before being first processed. This allows easier resumed restores.
 - New option: `no-require-git` for backup - if enabled, a git repository is not required to apply `git-ignore` rule.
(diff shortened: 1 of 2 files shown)


aawsome:
Thanks for the PR - I think this is a good improvment.

It seems to me that this PR doesn't change the wait-behavior. Instead it seem that now only stdout is used to read the password from while stderr is not catched anymore and stdin can now be used to manually input something into the command (instead of forcing the command to abort, which was the previous behavior). Is this correct?
Can you please also add a changelog entry?

istudyatuni:
> now only stdout is used to read the password from while stderr is not catched anymore and stdin can now be used to manually input something into the command (instead of forcing the command to abort, which was the previous behavior)

Yes. Also, stderr, for example in keepassxc-cli, is used to show password prompt.

I think maybe it's a good idea to exit, if command exited with non-zero code, will do

aawsome (approved):
LGTM! Thanks a lot again!

PR #702, opened 2023-06-23
allow request for new key to confirm entered pass

Aderemi-Adesada:
allow request for new key to prompt user to confirm the password they entered

### DIFF
--- a/src/commands/init.rs
+++ b/src/commands/init.rs
@@ -8,7 +8,7 @@ use anyhow::{bail, Result};
 use crate::{commands::get_repository, Application, RUSTIC_APP};
 
 use bytes::Bytes;
-use rpassword::prompt_password;
+use dialoguer::Password;
 
 use rustic_core::{
     hash, random_poly, ConfigFile, DecryptBackend, DecryptWriteBackend, FileType, Id, Key, KeyFile,
@@ -80,7 +80,11 @@ pub(crate) fn save_config(
     let key = Key::new();
 
     let pass = password.map_or_else(
-        || match prompt_password("enter password for new key: ") {
+        || match Password::new().with_prompt("enter password for new key")
+        .allow_empty_password(true)
+        .with_confirmation("confirm password", "passwords do not match")
+        .interact()
+        {
             Ok(it) => it,
             Err(err) => {
                 status_err!("{}", err);
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2233,17 +2233,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "rpassword"
-version = "7.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6678cf63ab3491898c0d021b493c94c9b221d91295294a2a5746eacbe5928322"
-dependencies = [
- "libc",
- "rtoolbox",
- "winapi",
-]
-
-[[package]]
 name = "rstest"
 version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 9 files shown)


istudyatuni:
Maybe move these closures to separate function?

Aderemi-Adesada:
What is the best file to add the function so that both key.rs and init.rs can access

aawsome:
First, this PR IMO doesn't allow to confirm the password, but enforces it.
I personally wouldn't need that, but if you think this is a worthwhile addition, I'm open for it ;-)

About the place: Please not in `rustic_core`. This crate provides core functionality to access repositories and perform operations on it. All UI topics should stay outside - and this is for sure just a UI topic!

Aderemi-Adesada:
@istudyatuni @aawsome  what do you think about having a function called prompt_password_twice in rustic/src/helpers that both init.rs and key.rs can call

istudyatuni:
I'm afraid to make those decisions in this project 😁

aawsome:
As discussed @Aderemi-Adesada on discord, I also think that the best idea is to replace `rpassword`  with `dialoguer` which already implements this functionality.

Aderemi-Adesada:
rpassword has been replaced with dialoguer in init.rs and key.rs when prompting for new key

src/commands/init.rs
```diff
@@ -80,12 +80,19 @@ pub(crate) fn save_config(
...
+                .interact();
+
+            let password = match password_result {
+                Ok(it) => it,
+                Err(err) => {
+                    status_err!("{}", err);
+                    RUSTIC_APP.shutdown(Shutdown::Crash);
+                }
+            };
+            password
```
aawsome, line 95: No need for new variables which are immediately used in the next command. You can use `match Password::new() ... { ... }` here.

src/commands/init.rs
```diff
@@ -80,12 +80,19 @@ pub(crate) fn save_config(
...
     let pass = password.map_or_else(
-        || match prompt_password("enter password for new key: ") {
-            Ok(it) => it,
-            Err(err) => {
-                status_err!("{}", err);
-                RUSTIC_APP.shutdown(Shutdown::Crash);
-            }
+        || -> String {
+            let password_result = Password::new().with_prompt("enter password for new key")
+                .with_confirmation("confirm password", "passwords do not match")
```
aawsome, line 85: According to the documentation you have to enable empty passwords explicitly. These are currently supported by rustic, so please add `allow_empty_password(true)`.

aawsome (changes requested):
Beside my remarks, I was wondering if we can remove the dependency on `rpassword` completely. Then I remembered some newline handling which needed to be dealt with.
Checking the code I realized that that fix already removed `rpassword` when reading from a file - but obviously I have overseen this in `src/command/key.rs`. Can you please have a look at `read_password_from_reader` in `rates/rustic_core/src/repository.rs` and use this in  `src/command/key.rs` - this should allow to remove `rpassword` completely.

Moreover, can you please add an entry in `changelog/new.txt` and after your changes squash all commits into 1 commit? Thanks a lot!

Aderemi-Adesada:
rpassword have been removed completely and `src/command/key.rs` now uses `read_password_from_reader` as used in `crates/rustic_core/src/repository.rs`

aawsome:
@Aderemi-Adesada there are syntax errors in `lib.rs` - can you have a look at it?

Aderemi-Adesada:
> syntax error

I will fix that right now

Aderemi-Adesada:
syntax error have been fixed

aawsome:
Can you please rebase to main and squash all commits instead of re-merging master back into your branch? That will keep commit history clean. Thanks a lot!

aawsome:
Oh, clicked wrong - actually wanted to run the tests.
Anyway, we'll see in the CI. Thanks a lot @Aderemi-Adesada - looks good to me!

Aderemi-Adesada:
You are welcome

PR #739, opened 2023-07-08
Systemd timers for backup and forget

nils-werner:
(Related to [this discussion](https://github.com/rustic-rs/rustic/discussions/609#discussioncomment-6262498))

One thing that I think is particularly handy with crestic and config profiles is the possibility of creating systemd timers that periodically run backups and forgets. In the case of the config files of this PR, backup would run daily, forget would run monthly.

Right now the files in the PR are not really installed anywhere because `cargo build` doesn't support creating datafiles or creating Linux packages, so I guess it'd be the task of the downstream packagers to put these files in the packages. There are two locations where these files could go:

 - `/usr/lib/systemd/system/` for system units (i.e. run by `root` and config files in `/etc/`)
 - `/usr/lib/systemd/user` for user units (i.e. run by each user and config files in `~/.config/`)

However since rustic currenly seems to only read config files from `~/.config/rustic` and not from `/etc/rustic`, only user units make sense. Each individual user can then activate one ore more timers using

```
systemctl --user activate --now rustic-backup@PROFILENAME.timer
```

to have the system periodically run

```
rustic --user-profile PROFILENAME backup
```

### DIFF
--- a/util/systemd/rustic-backup@.service
+++ b/util/systemd/rustic-backup@.service
@@ -1,0 +1,8 @@
+[Unit]
+Description=rustic --use-profile %I backup
+
+[Service]
+Nice=19
+IOSchedulingClass=idle
+KillSignal=SIGINT
+ExecStart=/usr/bin/rustic --use-profile %I backup
--- a/util/systemd/rustic-forget@.service
+++ b/util/systemd/rustic-forget@.service
@@ -1,0 +1,6 @@
+[Unit]
+Description=rustic --use-profile %I forget
+
+[Service]
+KillSignal=SIGINT
+ExecStart=/usr/bin/rustic --use-profile %I forget
--- a/util/systemd/rustic-backup@.timer
+++ b/util/systemd/rustic-backup@.timer
@@ -1,0 +1,12 @@
+[Unit]
+Description=Daily rustic --use-profile %I backup
+Wants=rustic-forget@%i.timer
+
+[Timer]
+OnCalendar=daily
+AccuracySec=1m
+RandomizedDelaySec=1h
+Persistent=true
+
+[Install]
+WantedBy=timers.target
--- a/util/systemd/rustic-forget@.timer
+++ b/util/systemd/rustic-forget@.timer
@@ -1,0 +1,12 @@
+[Unit]
+Description=Monthly rustic --use-profile %I forget
+PartOf=rustic-backup@%i.timer
+
+[Timer]
+OnCalendar=monthly
+AccuracySec=1m
+RandomizedDelaySec=1h
+Persistent=true
+
+[Install]
+WantedBy=timers.target


aawsome:
I'm out of home ATM, so just a comment: reading config files from /etc/ has been recently added, available in the latest beta versions.

Am 8. Juli 2023 09:13:28 MESZ schrieb Nils Werner ***@***.***>:
>(Related to [this discussion](https://github.com/rustic-rs/rustic/discussions/609#discussioncomment-6262498))
>
>One thing that I think is particularly handy with crestic and config profiles is the possibility of creating systemd timers that periodically run backups and forgets. In the case of the config files of this PR, backup would run daily, forget would run monthly.
>
>Right now the files in the PR are not really installed anywhere because `cargo build` doesn&#39;t support creating datafiles or creating Linux packages, so I guess it&#39;d be the task of the downstream packagers to put these files in the packages. There are two locations where these files could go:
>
> - `/usr/lib/systemd/system/` for system units (i.e. run by `root` and config files in `/etc/`)
> - `/usr/lib/systemd/user` for user units (i.e. run by each user and config files in `~/.config/`)
>
>However since rustic currenly seems to only read config files from `~/.config/rustic` and not from `/etc/rustic`, only user units make sense. Each individual user can then activate one ore more timers using
>
>```
>systemctl --user activate --now ***@***.***
>```
>
>to have the system periodically run
>
>```
>rustic --user-profile PROFILENAME backup
>```
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/rustic-rs/rustic/pull/739
>
>-- Commit Summary --
>
>  * Systemd timers for backup and forget
>
>-- File Changes --
>
>    A ***@***.*** (8)
>    A ***@***.*** (12)
>    A ***@***.*** (6)
>    A ***@***.*** (12)
>
>-- Patch Links --
>
>https://github.com/rustic-rs/rustic/pull/739.patch
>https://github.com/rustic-rs/rustic/pull/739.diff
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/rustic-rs/rustic/pull/739
>You are receiving this because you are subscribed to this thread.
>
>Message ID: ***@***.***&gt;

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

nils-werner:
> I'm out of home ATM, so just a comment: reading config files from /etc/ has been recently added, available in the latest beta versions.

Ah, good to know! In that case the files should go into both directories.

util/systemd/rustic-forget@.service
```diff
@@ -0,0 +1,6 @@
+[Unit]
+Description=rustic --use-profile %I forget
+
+[Service]
+KillSignal=SIGINT
+ExecStart=/usr/bin/rustic --use-profile %I forget
```
aawsome, line 6: I would add `--prune` here. Usually just forgetting without pruning doesn't help that much...

util/systemd/rustic-backup@.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=rustic --use-profile %I backup
+
+[Service]
+Nice=19
+IOSchedulingClass=idle
+KillSignal=SIGINT
+ExecStart=/usr/bin/rustic --use-profile %I backup
```
aawsome, line 8: What about adding a comment that this needs a config file `%I.toml` in`/etc/rustic/` or the users config dir `~/.config/rustic/`?

aawsome (changes requested):
Thank you very much @nils-werner for the PR!
I think it is very valuable having this!

util/systemd/rustic-forget@.service
nils-werner: Doesn't it make most sense to let people put that in their configs?

util/systemd/rustic-backup@.service
nils-werner: I'm not sure... Why put the comment here, shouldn't it be "general knowledge" that users have to have config files in order to use profiles?

util/systemd/rustic-forget@.service
aawsome: Ah, of course you are right. The `prune = true/false` option in the config wasn't always present, but now that it is, it makes much more sense putting it there.

util/systemd/rustic-backup@.service
aawsome: I thought more about the fact that people directly see the connection between a `rustic-backup@my-profile.service` and rustic using the profile `my-profile`.

nils-werner: It's kind of in the description already

```
Description=rustic --use-profile %I backup
```

And I think I'd rather put it in the documentation somewhere... The units need to be documented anyways, otherwise nobody will find them in the first place.

aawsome (approved):
LGTM! Thanks again @nils-werner

PR #740, opened 2023-07-08
Fix "forget" command not removing provided IDs' snapshots

Aderemi-Adesada:
Fix "forget" command not removing provided IDs' snapshots. Its says nothing to remove when there are snapshots matching  the id


aawsome (approved):
LGTM! Thanks for the bugfix!

PR #752, opened 2023-07-16
use shellwords to split strings

aawsome:


aawsome:
I came to shell_words by realizing that we already have it as dependency (in `rustic`, as `dialoguer` depends on it) and that it in fact solves the problem I tried to solve by the `nom` crate. It does so even better as the current solution doesn't support and handle `\"` in strings. So it is both an improvement and simplifies dependencies (a bit). 
I'm open to exchange it with something else if there is a better choice! If not, I would just use it as the very popular `dialoguer` crate also uses it...

PR #762, opened 2023-07-17
docs: Add `CONTRIBUTING.md`

### LINKED ISSUE #755: Add `CONTRIBUTING.md`
We should add a `CONTRIBUTING.md` containing a few guidelines and things people should know before contributing:

- Explaining issue labels WRT https://github.com/rust-lang/rust/labels?page=1&sort=name-asc and https://github.com/winnow-rs/winnow/labels
```
A - Area
C - Category
D - Diagnostic
E - Call for participation
F - Feature
I - Issue e.g. I-crash
M - Meta
O - Operating systems
P - priorities e.g. P-{low, medium, high, critical}
PG - Project Group
perf - Performance
S - Status e.g. S-{blocked, experimental, inactive}
T - Team relevancy
WG - Working group
```

- guideline should state clearly that code that is changed and PRed needs to be tested (so the old code needs to be covered, and the test needs to be passing for the new code as well)
  - test coverage currently is low but we need a higher test coverage to refactor stuff, so this guideline seems reasonable 
  - like this we raise continuously the test coverage and make sure we don't pull in logical bugs, while we can still accept PRs

- workflow for PRs
  - e.g. not force pushing, so we can keep review comments

- example `CONTRIBUTING.md`: https://github.com/SFTtech/openage/blob/master/doc/contributing.md

simonsan:
closes #755 

TODO:
- [x] add issue/PR labels and explanations
    ```
    A - Area
    C - Category
    D - Diagnostic
    E - Call for participation
    F - Feature
    I - Issue e.g. I-crash
    M - Meta
    O - Operating systems
    P - priorities e.g. P-{low, medium, high, critical}
    PG - Project Group
    perf - Performance
    S - Status e.g. S-{blocked, experimental, inactive}
    T - Team relevancy
    WG - Working group
    ```


CONTRIBUTING.md
```diff
@@ -0,0 +1,211 @@
...
+-->
+
+## License
+
+By contributing to `rustic` or any crates contained in this repository, you agree that your contributions will be licensed under either of:
+
+- [Apache License, Version 2.0](./LICENSE-APACHE)
+- [MIT license](./LICENSE-MIT)
+
+at your option. Unless you explicitly state otherwise, any contribution intentionally submitted
```
aawsome, line 209: I think that should be something like "at the option of everyone using `rustic`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,211 @@
...
+
+There's also `git commit --amend` which is a "mini-rebase" that modifies just the last commit with your current changes by `git add`.
+It just skips the creation of a new commit and instead melds the changes into the last one you made.
+
+If you want to update a single commit in the range `[upstream/master, current HEAD]` which is not the last commit:
+
+- `edit stuff you wanna change in some previous commit`
+- `git add changed_stuff`
+- `git commit --fixup $hash_of_commit_to_be_fixed`
+- `git rebase --autosquash -i -m upstream/master`
```
aawsome, line 170: Wow cool - didn't know about this possibility! Have to test it out! I so far used `git rebase -i` and manually resorted the fixups... Thanks for that hint!

aawsome (commented):
I like that document very much. I don't know if I can read the Rust CoC in near future, but I think we can start without it and work on that a little bit later.
But I agree we *should* work on this, as I like to have a clear policy about how people can join the project - and I would like to have a procedure to let people easily and transparently join.

CONTRIBUTING.md
aawsome: Oh, I somehow missed your question, sorry for that!
We want to have rustic dual-licenced which means that the *user* of rustic may choose which licence to use. That means, that we must require contriubtors to also have their work dual-licenced in the same sense. If we write "you agree that your contributions will be licensed under either of:

- [Apache License, Version 2.0](./LICENSE-APACHE)
- [MIT license](./LICENSE-MIT)

at your option." I think this can be misunderstanding that a contributor may choose the licence under which their work will be licenced. But this is not what we want.

aawsome (approved):
LGTM! Thanks a lot @simonsan

PR #766, opened 2023-07-18
docs: Fully document configuration specification and environment variables

### LINKED ISSUE #765: Fully document configuration specification
There are many values that are missing annotations in `config/*.toml` files. Inline comments are a good start. I think it would be valuable to document all available configuration options (the specification) in a table in a `Readme.md` in the `config/` directory so users can easily look up what can be set via the configuration files.

### LINKED ISSUE #768: Fully document environment variables
`rustic` reads various environment variables at runtime, that also interfere with configuration settings.

For a better overview, we should therefore document:
- [x] which environment variables there are
- [x] what is the priority these values are applied (`cli > env > config`?)

simonsan:
Fixes #765
Fixes #768

TODO:
- [x] add `example` column and split/move things from `default` values


config/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Global Options
+
+| Attribute         | Description                                            | Default Value  | Example Value | Environment Variable |
+|-------------------|--------------------------------------------------------|----------------|---------------|---------------|
+| dry-run           | If true, performs a dry run without making any changes. | false       || RUSTIC_DRY_RUN |
+| log-level         | Logging level. Possible values: "off", "error", "warn", "info", "debug", "trace". | "debug" || RUSTIC_LOG_LEVEL |
+| log-file          | Path to the log file.                                 | TODO | "/log/rustic.log" | RUSTIC_LOG_FILE |
+| no-progress       | If true, disables progress indicators.               | false          || RUSTIC_NO_PROGRESS |
+| progress-interval | The interval at which progress indicators are shown. | "100ms"        || RUSTIC_PROGRESS_INTERVAL |
```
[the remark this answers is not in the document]
aawsome: its both an example and the default value.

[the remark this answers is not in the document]
aawsome: But a better example would actually to have a much longer progress intervall like "1m"

config/README.md
```diff
@@ -0,0 +1,125 @@
...
+| use-profile       | An array of profiles to use.                           | Empty array    || RUSTIC_USE_PROFILE |
+
+### Repository Options
+
+| Attribute         | Description                                            | Default Value  | Example Value | Environment Variable |
+|-------------------|--------------------------------------------------------|----------------|---------------|---------------|
+| cache-dir         | Path to the cache directory.                          | TODO        | Default cache dir, e.g., ~/.cache/rustic | RUSTIC_CACHE_DIR |
```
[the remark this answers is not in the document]
aawsome: yes, exactly like it is written. But I think it will be the repository-depending dir, i.e. `~/.cache/rustic/$REPO_ID`, and when you specify another one it won't add the repository id, IIRC.

config/README.md
```diff
@@ -0,0 +1,125 @@
...
+| git-ignore             | If true, use .gitignore rules to exclude files from the backup in the source directory. | true ||
+| source                 | Source directory or file to back up.                  | Not set        ||
+
+### Forget Options
+
+| Attribute         | Description                                            | Default Value  | Example Value |
+|-------------------|--------------------------------------------------------|----------------|---------------|
+| filter-host            | Array of hosts to filter snapshots.                   | Not set        | ["forgethost"] |
+| keep-daily             | Number of daily backups to keep.                      | Not set        ||
+| keep-within-daily      | The time duration within which daily backups will be kept. | TODO      | "7 days"     |
```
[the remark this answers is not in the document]
aawsome: All keep-* options are not set by default.

config/README.md
```diff
@@ -0,0 +1,125 @@
...
+| keep-monthly           | Number of monthly backups to keep.                    | Not set        ||
+| keep-weekly            | Number of weekly backups to keep.                     | Not set        ||
+| keep-yearly            | Number of yearly backups to keep.                     | Not set        ||
+| keep-tags              | Array of tags to keep.                                | Not set        | ["mytag"]      |
+
+### Copy Targets
+
+| Attribute         | Description                                            | Default Value  | Example Value |
+|-------------------|--------------------------------------------------------|----------------|---------------|
+| cache-dir              | Path to the cache directory for the target repository. | TODO | Default cache dir, e.g., ~/.cache/rustic |
```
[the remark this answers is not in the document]
aawsome: yes. Copy-targets are simply repositories with the same defaults as within the repository section.

config/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+    Commandline Arguments >> Environment Variables >> Configuration File
+
+Values parsed from the `configuration file` can be overwritten by `environment variables`, which can be overwritten by `commandline arguments` options. Therefore `commandline arguments` have the highest precedence.
+
+**NOTE**: There are the following restrictions:
+
+- You can overwrite values, but for most values, you cannot "unset" them on a higher priority level.
+
+- For some integer values, you cannot even overwrite with the value `0`, e.g. `keep-weekly = 5` in the `[forget]` section of the config file cannot be overwritten by `--keep-weekly 5`.
```
aawsome, line 19: should be: "cannot be overwritten by --- = 0"

config/README.md
```diff
@@ -0,0 +1,125 @@
...
+This is not relevant for env variables, only for some values available in the `config` and as `CLI` option.
```
aawsome, line 20: This is not fully correct - maybe just remove that sentence, IMO it is a bit distracting.

config/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Sections and Attributes
+
+### Global Options
+
+| Attribute         | Description                                            | Default Value  | Example Value | Environment Variable |
+|-------------------|--------------------------------------------------------|----------------|---------------|---------------|
+| dry-run           | If true, performs a dry run without making any changes. | false       || RUSTIC_DRY_RUN |
+| log-level         | Logging level. Possible values: "off", "error", "warn", "info", "debug", "trace". | "debug" || RUSTIC_LOG_LEVEL |
```
aawsome, line 29: default is info

config/README.md
```diff
@@ -0,0 +1,125 @@
...
+### Backup Options
+
+| Attribute         | Description                                            | Default Value  | Example Value |
+|-------------------|--------------------------------------------------------|----------------|---------------|
+| as-path                | Specifies the path for the backup when the source contains a single path. | Not set (Source-specific option) ||
+| description            | Description for the backup.                           | Not set        ||
+| description-from       | Path to a file containing the description for the backup. | Not set    ||
+| delete-never           | If true, never delete the backup.                     | false          ||
+| delete-after           | Time duration after which the backup will be deleted. | Not set        ||
+| exclude-if-present     | Array of filenames to exclude from the backup if they are present. | Not set (Source-specific option) ||
```
aawsome, line 76: Actually this and the following options are not source-specific, but if set here, they apply for all sources (but can be overwritten in the source-specifc configuration, see below)

config/README.md
```diff
@@ -0,0 +1,125 @@
...
+| stdin-filename         | File name to be used when reading from stdin.         | Not set (Source-specific option) ||
+| tag                    | Array of tags for the backup.                         | Not set        ||
+| with-atime             | If true, includes file access time (atime) in the backup. | Not set (Source-specific option) ||
+
+### Backup Sources
+
+| Attribute         | Description                                            | Default Value  | Example Value |
+|-------------------|--------------------------------------------------------|----------------|---------------|
+| git-ignore             | If true, use .gitignore rules to exclude files from the backup in the source directory. | true ||
+| source                 | Source directory or file to back up.                  | Not set        ||
```
aawsome, line 95: Maybe we should add here, that the option given here for the source are used, when the backup command is used to backup this specific source.

config/README.md
aawsome: I would prefer an intro text, actually...

config/README.md
aawsome: e.g. `keep-weekly = 5` in the `[forget]` section of the config file cannot be overwritten by `--keep-weekly 0`.

config/README.md
aawsome: Yes, I would remove it and add an intro text like "The following options can be specified for each source individually in the source-individual section, see below. If they are specified here, they provide default values for all sources but can still be overwritten in the source-individual section"

aawsome (approved):
LGTM. Thanks a lot!

PR #773, opened 2023-07-19
ci: Ignore markdown files in CI

simonsan:
For the CI, I think we should add some check if the PR only changes docs and markdown files, so it doesn't run all the (expensive) tests for the binary. It's a waste of CI time IMHO and thus also energy. #SaveThePlanet

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,7 +7,7 @@ on:
   push:
     branches:
       - main
-      - release
+      - release/*
     paths-ignore:
       - "**/*.md"
 


.github/workflows/release.yaml
```diff
@@ -1,5 +1,9 @@
 on:
   push:
+    branches:
+      - main
```
aawsome, line 4: Actually, the `release` pipeline is currently the only one to build rustic for all supported targets. If we introduce some platform-specific thing that breaks a specific target, it would fail in this pipeline. Therefore I would remove the "push-on-main-only" or replace it with "on pull-request-to-main"

aawsome (changes requested):
Mostly fine and a good idea!

.github/workflows/release.yaml
aawsome: Ah, yes. You are right!

aawsome (approved):
LGTM! Thanks a lot!

PR #785, opened 2023-07-23
handle non-unicode link-targets

### LINKED ISSUE #117: non-unicode link targets are not supported
I create this as a new issue so it will be still visible after #57 is closed.

This is so far not supported in restic, see https://github.com/restic/restic/issues/3311

So, this can be only fixed after restic fixes it or by breaking restic compatibility.
Until this situation is cleared out, this will get a wontfix label.

aawsome:
After this has been added to the restic repository format (https://github.com/restic/restic/pull/3802), this can be also handled within rustic.

Note that for windows, this PR assumes that there are no non-unicode link-targets.

closes #117


aawsome:
The restic PR has just been released in restic 0.16.0 - so we'll merge this soon, too!

aawsome (commented):
LGTM

PR #821, opened 2023-08-17
ci: Fail fast if something is not working out in build matrix

simonsan:
I think that this would be better for the contingent of the CI hours (I think it's 2000 hours per month). If something doesn't work for one platform, we rather fix it anyway and don't release binaries for other platforms, as we can't be sure, that they won't have unintended side effects either?


aawsome:
According to https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast fail-fast is enabled by default. This is also the behavior I already observed with the current CI-pipeline...

PR #829, opened 2023-08-18
chore: add support for `cargo-binstall`

simonsan:
After merge, we can add that to the installation section's `From binaries` in #820


aawsome (approved):
LGTM

PR #830, opened 2023-08-18
ci: Lint markdown with dprint, run initial dprint fmt

simonsan:
For our documentation efforts within the repository, it makes sense to introduce a common style that is easy to uphold using `dprint`. Also, because `dprint` is just a `cargo install dprint`/`scoop install dprint` away.


aawsome (approved):
LGTM.
The only thing is that the gh-actions plugin seems to be not very popular. Hope it keeps working as it should!
Thanks a lot @simonsan

PR #836, opened 2023-08-19
chore: Downgrade serde due to distributing binary blob 

simonsan:
Downgrade `serde` due to distributing binary blob for `serde_derive` for now until further steps have been figured out.

Ref.: `https://github.com/serde-rs/serde/issues/2538` (no link)

It could impose a security threat, and I'm actually unsure if that isn't even against the ToS/Guidelines on crates.io. CC: `https://github.com/rust-lang/crates.io/issues/6992` (no link)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,9 +43,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.19.0"
+version = "0.20.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


aawsome:
Just spend a little time to value the discussions on github and reddit.
I support to downgrade. Actually, only the `rustic` CLI depends on a specific `serde`/`serde_derive` version via the lock file. So maybe it is sufficient to just set the version in the lock file?
I'm not sure, however, which versions are taken when you run `cargo test`...

PR #843, opened 2023-08-22
ci: fix artifacts not building, do non-cross builds with `cargo-auditable` and include `config/` directory into package

### LINKED ISSUE #819: Embed auditable information about dependencies in compiled executable
I think it would be nice to embed the exact crate versions used to build our Rust executables. To be able to audit binaries for known bugs or security vulnerabilities in production. [cargo-auditable](https://github.com/rust-secure-code/cargo-auditable) does this by embedding data about the dependency tree in JSON format into a dedicated linker section of the compiled executable. Afterwards, everyone is able to audit it with [cargo-audit](https://crates.io/crates/cargo-audit) with the `--bin` flag.

screenshot-bin 1

simonsan:
A while ago there was the incentive by me to use [cargo-auditable](https://github.com/rust-secure-code/cargo-auditable) to embed auditable metadata about used dependencies. We haven't followed that up, because it was breaking our `cross` builds.

With this PR we embed auditable dependency metadata only for non-cross builds and also package the `config/` directory to make it easier to start out with `rustic`.

fixes #819

This also fixes a bug with the packaging within the build workflow for PRs. The problem was, that Github probably changed their internal branch naming for `github.ref_name` to something like `branch_bla/843/merge`, so the packaging step was not running anymore and the following actions were continuing without failing but the outcome were non-existing artifacts.

## Artifacts are built again
2023-08-22 07_45_05-WDADesktopService

## Hashfile is being generated as well
2023-08-22 07_45_56-C__Users_dailyuse_Downloads_binary-armv7-unknown-linux-gnueabihf zip_

## Config files are being packaged as well
2023-08-22 07_46_33-C__Users_dailyuse_Downloads_binary-armv7-unknown-linux-gnueabihf zip_rustic-5934

## Dependency metadata is embedded and readable

2023-08-22 07_55_29-Settings


aawsome:
IIRC, the hash files broke `self-update` which didn't work when the hash files were also present in the releases. The other changes are ok, IMO.

PR #845, opened 2023-08-23
feat: Securely zero memory for Passwords using secrecy `Zeroize` trait

simonsan:
Started to use `RusticPassword` where applicable.

@aawsome Could you let me know where else it would be good as well?

`rustic_core::crypto::Key` I'm aware of and currently working on. It doesn't look as straight forward as the password, though, due to more generic bounds. I will try doing that in the following commits.

Should fix rustic-rs/rustic_core#9

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2507,9 +2507,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-webpki"
-version = "0.100.1"
+version = "0.100.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


aawsome:
So, basically there is the password used by end-users (either given directly, in a file or via a command). This is then used (in combination with a key derivation function) to generate a decryption key for the information given in the `KeyFile`. The decrypted information is then again the master key for the repository. So, it is the password, the derived key and the master key.

I also think that `Zeroize` must also be used within the `aes256ctr_poly1305aes` crate to cover all occurences of key used in rustic.

PR #871, opened 2023-09-08
Feature: Add s3 backend

### LINKED ISSUE #546: add native s3 storage backend
I was having a look at implementing it myself, but found the backend traits to not be async.  I also found in the changelog, that async was removed in 0.4.  Is this going to change soonish again?  If so, I'll defer it.

I would use `rust-s3`, because afaict the official aws sdk is async only.  What do you think?

Scandiravian:
This PR adds an S3 backend which improves the drop-in-ability of rustic as a replacement for Restic.

When reviewing the code I highly recommend doing it one commit at a time, since some of these will be dropped before the PR is marked as ready. I have included them to show how I have tested the code.

When the PR is ready for review I will also squash the undropped commits into a single one, as per the contribution guidelines.

closes #546

### DIFF
--- a/crates/rustic_core/src/backend/s3.rs
+++ b/crates/rustic_core/src/backend/s3.rs
@@ -1,0 +1,355 @@
+use std::time::Duration;
+use std::{collections::HashSet, env};
+
+use crate::{
+    backend::{FileType, ReadBackend, WriteBackend, ALL_FILE_TYPES},
+    error::S3ErrorKind,
+    Id, RusticResult,
+};
+
+use backoff::{backoff::Backoff, ExponentialBackoff, ExponentialBackoffBuilder};
+use log::{info, trace, warn};
+use s3::{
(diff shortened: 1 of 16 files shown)


crates/rustic_core/src/backend/s3.rs
```diff
@@ -0,0 +1,406 @@
...
+impl ReadBackend for S3Backend {
+    fn location(&self) -> String {
+        let mut location = "s3:".to_string();
+        location.push_str(&self.path_style_url());
+        location
+    }
+
+    fn set_option(&mut self, _option: &str, _value: &str) -> RusticResult<()> {
+        todo!()
+    }
```
Scandiravian, line 224: I'm not sure what should be implemented here, so some help on this would be really appreciated

Scandiravian:
I've been somewhat busy the last two weeks, so haven't had time to move the PR yet. Will try to get to it during October :smile:

PR #54, opened 2022-07-03
Add jemalloc as allocator for supported platforms

aawsome:
maybe this helps with #48

PR #897, opened 2023-09-19
feat(alloc): Add optional (unsupported) allocators

simonsan:
Combines #54 and #229 and introduces mutually exclusive features to enable them.


aawsome:
Actually, I wouldn't blow up the CI tripling many tests for adding (unsupported) features of enabling a global allocator. If there are problems with the allocator, it should be allocators business and elsewise I expect all tests to run as well.

aawsome (approved):
LGTM!

PR #906, opened 2023-10-12
init: Handle option --dry-run

### LINKED ISSUE #905: `rustic init --dry-run` creates repo
But help says:
```
--dry-run  -- Only show what would be done without modifying anything. Does not affect read-only commands
```

Seems it ignores `--dry-run`.

_(`rustic --version` =  `0.6.0`, built locally from 54c608c)_

aawsome:
fixes #905

### DIFF
--- a/src/commands/init.rs
+++ b/src/commands/init.rs
@@ -43,6 +43,14 @@ impl InitCmd {
             bail!("Config file already exists. Aborting.");
         }
 
+        // Handle dry-run mode
+        if config.global.dry_run {
+            bail!(
+                "cannot initialize repository {} in dry-run mode!",
+                repo.name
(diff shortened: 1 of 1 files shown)


zxcat (approved):
For some reason I have a right to review. PR seems ok for me.
The only thing to comment: afaiu this change makes `--dry-run` invalid to use with `init`. But it's also possible to allow its usage in a normal way: just skip any init actions, but check if repo creation will be ok (at least until something written), also stopping rustic with `0` exit code. In this case "the spirit" of `--dry-run` is still there: it may be used to check if current config + passed arguments are ok before doing something.
But the current solution is simple enough, idk is it worth to change anything

PR #913, opened 2023-10-19
Fix term log level limiter when `log-file` option set

zxcat:
referenced here: https://github.com/rustic-rs/rustic/discussions/907#discussioncomment-7264998


aawsome (approved):
Good catch! Thanks for the PR

PR #941, opened 2023-11-14
backup: Add option to omit identical backups

aawsome:
Once https://github.com/rustic-rs/rustic_core/pull/56 is merged and released, this PR is needed to add the option to rustic.


aawsome:
After https://github.com/rustic-rs/rustic_core/pull/90, the option should be automatical available after bumping rustic_core

PR #969, opened 2023-12-15
Add Dockerfile

Vatson112:
Hello, I decided to test rustic in my environment. However I see that you dont have docker image. So I decided to promose mine.

I'm new to the rust environment, so there are few questions:

- I dont propose any environment for your CI for autopublishing to docker.hub. I can take a look at this if you are interested.
- I checked rust-awesome, and there was mention about Alpine image and MUSL performance. So I used debian as runtime image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,27 @@
+# Improve build speed with cached deps
+ARG RUST_VERSION=1.70.0
+FROM lukemathwalker/cargo-chef:latest-rust-${RUST_VERSION} AS chef
+WORKDIR /app
+
+FROM chef AS planner
+COPY . .
+RUN cargo chef prepare --recipe-path recipe.json
+
+FROM chef AS builder
+COPY --from=planner /app/recipe.json recipe.json
+# Build dependencies - this is the caching Docker layer!
+RUN cargo chef cook --release --recipe-path recipe.json
+# Build application
+COPY . .
+RUN cargo build --release
+
+# why we dont use alpine for base image - https://andygrove.io/2020/05/why-musl-extremely-slow/
+FROM debian:bookworm as runtime
+
+COPY --from=builder /app/target/release/rustic /usr/local/bin
+
+# usually container will be used with --rm option
+# so we ignore cache
+ENV RUSTIC_NO_CACHE=true
+
+ENTRYPOINT ["/usr/local/bin/rustic"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,12 +16,8 @@ COPY . .
 RUN cargo build --release
 
 # why we dont use alpine for base image - https://andygrove.io/2020/05/why-musl-extremely-slow/
-FROM debian:bookworm as runtime
+FROM debian:bookworm-slim as runtime
 
 COPY --from=builder /app/target/release/rustic /usr/local/bin
-
-# usually container will be used with --rm option
-# so we ignore cache
-ENV RUSTIC_NO_CACHE=true
 
 ENTRYPOINT ["/usr/local/bin/rustic"]


aawsome:
Thanks for the PR @Vatson112! Of course we can publish a Dockerfile together with rustic!

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN cargo build --release
+
+# why we dont use alpine for base image - https://andygrove.io/2020/05/why-musl-extremely-slow/
+FROM debian:bookworm as runtime
+
+COPY --from=builder /app/target/release/rustic /usr/local/bin
+
+# usually container will be used with --rm option
+# so we ignore cache
+ENV RUSTIC_NO_CACHE=true
```
aawsome, line 25: Just wanted to note that using remote repositories without a cache can be very slow...
For many commands even using a cache which is directly removed after execution of the binary is in this case even faster than using no cache at all. So I would remove this, actually.

aawsome (changes requested):

aawsome:
> If there is any use case where it can be useful for running rustic in a container, please let me know, though.

I think one use case is people distributing software as docker images to be run onto their systems. Having rustic in an image then would be the identical way of distribution.

Vatson112:
> Nice, let me test that with [`podman`](https://podman.io/) the next days.

I also build and use it using pomdan. But I also tested it on Docker.

There are few useful examples how I used it.

Export vars for repo

```bash
export RUSTIC_REPOSITORY=/var/backup
export RUSTIC_PASSWORD=rustic
```

Get snapshots

```bash
podman run --rm -it -e RUSTIC_REPOSITORY=$RUSTIC_REPOSITORY -e RUSTIC_PASSWORD=$RUSTIC_PASSWORD   -v /var/backup:/var/backup:Z rustic  snapshots
```

Backup pg database

```bash
podman exec db pg_dumpall -U postgres | podman run --rm -i -v /var/backup:/var/backup:Z -e RUSTIC_REPOSITORY=$RUSTIC_REPOSITORY -e RUSTIC_PASSWORD=$RUSTIC_PASSWORD rustic backup --stdin-filename backup.sql -
```

> Or what would be your use case for that?

I don't like distributing software to servers as a binary. And it took me longer to build the RPM package than the container image. So I just took faster way.

It also occurs to me that the container can be used in immutable operating systems (like FCOS), and perhaps in a kubernetes cluster (if using Velero is impossible for some reason).

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+FROM chef AS builder
+COPY --from=planner /app/recipe.json recipe.json
+# Build dependencies - this is the caching Docker layer!
+RUN cargo chef cook --release --recipe-path recipe.json
+# Build application
+COPY . .
+RUN cargo build --release
+
+# why we dont use alpine for base image - https://andygrove.io/2020/05/why-musl-extremely-slow/
+FROM debian:bookworm as runtime
```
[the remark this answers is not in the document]
Vatson112: Nope, I will change it with next commit.

Dockerfile
Vatson112: Okey. Also remove it with next commit

aawsome (approved):
LGTM

aawsome:
Thanks @Vatson112

Vatson112:
Hi @simonsan 

I added basic docs. Check it out - https://github.com/rustic-rs/dev-docs/pull/13

Also what do you think about distribute `rustic` as container image? I checked your release docs and as I understood for stable release developer build and push release to `Crates/Github` manualy. Am I right?

mueckinger:
> In general, I would say, that containerization is useful for rustic only for making the build process easier and thus development, maybe (although `docker`/containerization is a pretty hard dependency to build on, so I wouldn't double down on that). If there is any use case where it can be useful for running rustic in a container, please let me know, though.

I am an experienced Kubernetes DevOps engineer and we heavily use restic containers for backing up everything in our clusters. I have my eye on rustic because of features like the hot/cold storage pattern and the backend abstraction via rclone. But not having an official and maintained docker image is a showstopper for rustic's usage in cloud environments.

BTW if you dislike dockerhub, github runs it's own container registry at ghcr.io . Don't know if it's free for open-source projects and how easily it can be integrated in your github CI/CD workflows. But if you want this project to be found by a broader audience, dockerhub would probably be the place to go.

mueckinger:
Sure, I can support with documentation and examples for usage with Kubernetes.

@Vatson112: Do you have experience in automated building and publishing to dockerhub or ghcr.io?

Just for my understanding:
From the code in rustic_core I can see that you are spawning an `rclone serve restic` instance, when using the rclone backend. So we can have a separate rclone container running `rclone serve restic` to which we can communicate via REST API? This would be perfect considering the microservice architecture.

Vatson112:
Hi @mueckinger

> @Vatson112: Do you have experience in automated building and publishing to dockerhub or ghcr.io?

Nope. I used only selfhosted Harbor or Quay with Gitlab. 

But I think I can try it on next week. It`s unlikely there`s any rocket science here.

neurekadev:
Do we have a docker image readily available?

> In general, I would say, that containerization is useful for rustic only for making the build process easier
@simonsan I like using docker for server backups because I can define a docker compose file and run it anywhere.

By containerizing everything, migration to a new server is often as simple as running bringing down the docker daemon, running rsync to transfer everything and then running `docker compose up -d`.

mueckinger:
docker pull ghcr.io/mueckinger/rustic:0.7.0

PR #973, opened 2023-12-26
feat(commands): Add `mount` command

### LINKED ISSUE #971: `mount` command
There is [fuse-rs](https://github.com/zargony/fuse-rs) and this is one of the few missing features to feature parity with Restic. I sometimes find it very handy to just browse a snapshot with all my available tools.

aawsome:
This PR adds the `mount` command to rustic to access snapshot contents as a read-only filesystem when the feature-flag `mount` is chosen. 

As for the `webdav` command, there are following options:
- Access to a concrete snapshot/path, e.g. `rustic mount /mnt 37a63e5b:/my/path`.
- Access to all snapshots (maybe restricted by filters) by using templates to define a virtual directory structure where snapshots are located. Example: `rustic mount /mnt --path-template "[{hostname}]/[{label}]/{time}" --time-template "%Y-%m-%d_%H-%M-%S"` (these are also defined as default). Note that for all dirs containing only snapshots, also a `latest` entry is generated. `latest` and identical subsequent snapshots are symlinks when using `mount`.

This PR uses fuse_mt which is not optimal as it introduces some overhead (e.g. needs to save whole Paths in-memory).

Note: Requires https://github.com/rustic-rs/rustic_core/pull/331 to properly show all data of files (without it builds and runs, but files are truncated and return error when reading).

Note: Building with the `mount` feature flag requires special dependencies and is only possible on supported platforms, see https://github.com/cberner/fuser

closes #971

### DIFF
--- a/src/commands/mount.rs
+++ b/src/commands/mount.rs
@@ -1,4 +1,8 @@
 //! `mount` subcommand
+
+// ignore markdown clippy lints as we use doc-comments to generate clap help texts
+#![allow(clippy::doc_markdown)]
+
 mod fusefs;
 use fusefs::FuseFS;
 
@@ -10,7 +14,7 @@ use abscissa_core::{config::Override, Command, FrameworkError, Runnable, Shutdow
 use anyhow::{anyhow, Result};
 use conflate::Merge;
 use fuse_mt::{mount, FuseMT};
-use rustic_core::vfs::{IdenticalSnapshot, Latest, Vfs};
+use rustic_core::vfs::{FilePolicy, IdenticalSnapshot, Latest, Vfs};
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Command, Default, Debug, clap::Parser, Serialize, Deserialize, Merge)]
@@ -30,6 +34,11 @@ pub struct MountCmd {
     #[clap(long)]
     #[merge(strategy=conflate::bool::overwrite_false)]
     no_allow_other: bool,
+
+    /// How to handle access to files. [default: "forbidden" for hot/cold repositories, else "read"]
+    #[clap(long)]
+    #[merge(strategy=conflate::option::overwrite_none)]
+    file_access: Option<String>,
 
     /// The mount point to use
     #[clap(value_name = "PATH")]
@@ -120,7 +129,18 @@ impl MountCmd {
             ]);
         }
 
-        let fs = FuseMT::new(FuseFS::new(repo, vfs), 1);
+        let file_access = config.mount.file_access.as_ref().map_or_else(
+            || {
+                if repo.config().is_hot == Some(true) {
+                    Ok(FilePolicy::Forbidden)
+                } else {
+                    Ok(FilePolicy::Read)
+                }
+            },
+            |s| s.parse(),
+        )?;
+
+        let fs = FuseMT::new(FuseFS::new(repo, vfs, file_access), 1);
         mount(fs, mountpoint, &options)?;
 
         Ok(())
--- a/src/commands/mount.rs
+++ b/src/commands/mount.rs
@@ -1,38 +1,58 @@
 //! `mount` subcommand
 mod fusefs;
+use fusefs::FuseFS;
 
 use std::{ffi::OsStr, path::PathBuf};
 
-use crate::{repository::CliIndexedRepo, status_err, Application, RUSTIC_APP};
+use crate::{repository::CliIndexedRepo, status_err, Application, RusticConfig, RUSTIC_APP};
 
-use abscissa_core::{Command, Runnable, Shutdown};
-use anyhow::Result;
+use abscissa_core::{config::Override, Command, FrameworkError, Runnable, Shutdown};
+use anyhow::{anyhow, Result};
+use conflate::Merge;
 use fuse_mt::{mount, FuseMT};
 use rustic_core::vfs::{IdenticalSnapshot, Latest, Vfs};
+use serde::{Deserialize, Serialize};
 
-use fusefs::FuseFS;
-
-#[derive(clap::Parser, Command, Debug)]
-pub(crate) struct MountCmd {
+#[derive(Clone, Command, Default, Debug, clap::Parser, Serialize, Deserialize, Merge)]
+#[serde(default, rename_all = "kebab-case", deny_unknown_fields)]
+pub struct MountCmd {
     /// The path template to use for snapshots. {id}, {id_long}, {time}, {username}, {hostname}, {label}, {tags}, {backup_start}, {backup_end} are replaced. [default: "[{hostname}]/[{label}]/{time}"]
     #[clap(long)]
+    #[merge(strategy=conflate::option::overwrite_none)]
     path_template: Option<String>,
 
     /// The time template to use to display times in the path template. See https://docs.rs/chrono/latest/chrono/format/strftime/index.html for format options. [default: "%Y-%m-%d_%H-%M-%S"]
     #[clap(long)]
+    #[merge(strategy=conflate::option::overwrite_none)]
     time_template: Option<String>,
-
-    #[clap(value_name = "PATH")]
-    /// The mount point to use
-    mountpoint: PathBuf,
-
-    /// Specify directly which path to mount
-    #[clap(value_name = "SNAPSHOT[:PATH]")]
-    snap: Option<String>,
 
     /// Don't allow other users to access the mount point
     #[clap(long)]
+    #[merge(strategy=conflate::bool::overwrite_false)]
     no_allow_other: bool,
+
+    /// The mount point to use
+    #[clap(value_name = "PATH")]
+    #[merge(strategy=conflate::option::overwrite_none)]
+    mountpoint: Option<PathBuf>,
+
+    /// Specify directly which path to mount
+    #[clap(value_name = "SNAPSHOT[:PATH]")]
+    #[merge(strategy=conflate::option::overwrite_none)]
+    snapshot: Option<String>,
+}
+
+impl Override<RusticConfig> for MountCmd {
+    // Process the given command line options, overriding settings from
+    // a configuration file using explicit flags taken from command-line
+    // arguments.
+    fn override_config(&self, mut config: RusticConfig) -> Result<RusticConfig, FrameworkError> {
+        let mut self_config = self.clone();
+        // merge "webdav" section from config file, if given
+        self_config.merge(config.mount);
+        config.mount = self_config;
+        Ok(config)
+    }
 }
 
 impl Runnable for MountCmd {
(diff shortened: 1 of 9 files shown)


src/commands/mount/fs.rs
```diff
@@ -0,0 +1,222 @@
...
+        } else {
+            Err(libc::ENOSYS)
+        }
+    }
+
+    fn open(&self, _req: RequestInfo, path: &Path, _flags: u32) -> ResultOpen {
+        let node = self.node_from_path(path)?;
+        let open = self.repo.open_file(&node).map_err(|_| libc::ENOSYS)?;
+        let mut open_files = self.open_files.write().unwrap();
+        let fh = open_files.first_key_value().map(|(fh, _)| *fh).unwrap_or(0);
```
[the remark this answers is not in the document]
aawsome: Thanks, you are right!

src/commands/mount.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl MountCmd {
+    fn inner_run(&self) -> Result<()> {
+        let config = RUSTIC_APP.config();
+
+        let repo = open_repository(&config)?.to_indexed()?;
+        let node =
+            repo.node_from_snapshot_path(&self.snap, |sn| config.snapshot_filter.matches(sn))?;
+
+        let options = [OsStr::new("-o"), OsStr::new("fsname=rusticfs")];
```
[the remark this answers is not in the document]
aawsome: Sure. Also if there are sensible defaults, we should try to use them. What are good defaults in you opinion?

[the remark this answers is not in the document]
jjedelsky: Imho, `kernel_cache`, `allow_other`, and `default_permissions` are going to be the right choice most of the time. 

(For example, without `allow_other`, a fs mounted in [toolbox](https://github.com/containers/toolbox) (or any other container) would appear empty in gui file manager running outside of the container) 

It probably might be sometimes useful to disable `allow_other` and `default_permissions`.

src/commands/mount/fs.rs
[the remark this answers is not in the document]
aawsome: I think that the virtual fs (fs.rs) should go to rustic_core. It is used for both `mount` and `webdav` and the current place is wrong.

About the Filesystem implementations (mountfs.rs and webdavfs.rs): The problem is that they depend on the crates which also implement the serving of those FSs, fuse_mt and dav_handler. If we move this to rustic_core I would put them behind feature flags. But yes, I also tend to move these implementations to rustic core.

But I'll wait for the backend refactor PR before processing with the transfer ;-)

src/commands/mount.rs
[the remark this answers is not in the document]
aawsome: I'll add `kernel_cache` to the options.

I am wondering if we should just add an option like `_no_allow_other` to the `mount` command which then unsets both `allow_other` and `default_permission` - and without this option just set both...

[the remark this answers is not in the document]
jjedelsky: That sounds good.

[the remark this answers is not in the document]
aawsome: is added.

aawsome:
This PR is now ready but we may want to add the feature flag in our build-pipelines on selected targets (and include the dependencies) - in another PR.

src/commands/mount.rs
```diff
@@ -0,0 +1,101 @@
...
+        let config = RUSTIC_APP.config();
+
+        let path_template = self
+            .path_template
+            .clone()
+            .unwrap_or_else(|| "[{hostname}]/[{label}]/{time}".to_string());
+        let time_template = self
+            .time_template
+            .clone()
+            .unwrap_or_else(|| "%Y-%m-%d_%H-%M-%S".to_string());
```
simonsan, line 62: Is that something we want to be able to override from config?

src/commands/mount.rs
```diff
@@ -0,0 +1,101 @@
...
+use std::{ffi::OsStr, path::PathBuf};
+
+use crate::{repository::CliIndexedRepo, status_err, Application, RUSTIC_APP};
+
+use abscissa_core::{Command, Runnable, Shutdown};
+use anyhow::Result;
+use fuse_mt::{mount, FuseMT};
+use rustic_core::vfs::{IdenticalSnapshot, Latest, Vfs};
+
+use fusefs::FuseFS;
```
simonsan, line 13:
```suggestion
use crate::commands::mount::fusefs::FuseFS;
```

then we can merge it with the crate import on top as well

src/commands/mount/fusefs.rs
simonsan, line 1: Why do we need to have this in `rustic-rs` instead of `rustic_core`?

src/commands/mount/fusefs.rs
```diff
@@ -0,0 +1,235 @@
+use super::Vfs;
```
simonsan, line 1: Relative import

src/commands/mount/fusefs.rs
simonsan: Answered here: https://github.com/rustic-rs/rustic_core/pull/156#issuecomment-2410669666

src/commands/mount/fusefs.rs
aawsome: is fixed.

src/commands/mount.rs
aawsome: Is added. I just wonder how we want to use this in our config tests.. Should we add this to `full.toml` and the `check-config` fixture?

BTW: We have the same topic with the "webdav" feature, but this we include in all our tests. Here, we have a feature which is definitively not included in some targets, like windows...

Or maybe it is a better idea to compile the config structure unconditionally and just have the command conditionally compiled?

src/commands/mount.rs
aawsome: Actually I don't see why this should be changed - having an import of a submodule is quite standard, IMO. I just moved it up to the `use` statement.

simonsan: Just for consistency to always import from the crate root, because we do that everywhere I think

src/commands/mount.rs
simonsan: Hmm, the question is what happens if a config value is given on a system that doesn't support this, e.g. webdav or mount. Is there an error like 'The webdav feature is not supported on your system, it will be ignored in the config.'or something? Because that would be useful I guess and good UX.

aawsome: You get an error when parsing the config, e.g. when compiled without the `mount` feature:
```
error: rustic-rs fatal error: parse error: unknown field `mount`, expected one of `global`, `repository`, `snapshot-filter`, `backup`, `copy`, `forget`, `webdav` at line 32 column 1
```

I actually think we should exactly use this behavior. If there is no `mount` it is simply not wanted in the config. 

About tests I think we should add this to the `full.toml` and `show-config` tests - as soon as we add the `mount` option to the tests... (maybe this requires to use special test fixtures for windows...)

simonsan: Do we have control over the error message? Because that one is pretty bad and doesn't say anything what is happening and why this is an unknown field in this case. We need to document this error somewhere, so we don't possibly get a lot of issues about this.

Also, I'm not sure about the behaviour itself. Someone could migrate their config to another OS. E.g. from Linux to another Mac or Windows system, where they wouldn't be currently able to use mount. And then suddenly they get this error. Is there a way to check if the error is regarding `mount` or `webdav` and if so, spit out a specialized message, in regard to the missing feature on that platform or so?

aawsome: Mhh, ok, if we want to support copying tomls over, I agree that shouldn't be an error...
But then I would not even print a warning - if we use another command when having compiled with the `mount` option, that part of the config is also not used and we don't warn about that..

simonsan: I think we need to decide if the profile is OS-dependent or not. I would root for `not`, because it makes it easier to test as well as writing examples for the whole user base.

Instead of `deny_unknown_fields` I would propose giving feedback to the user, that certain keys are in the profile and not being used for whatever reason. e.g. `mount` and `webdav` if support is not compiled in, or if keys are used, that are not valid in a profile context.

I think that is pretty standard procedure to warn about content in a profile that has no effect, to not leave the user in the dark about it. As they may think, that something is set, and derive the current behaviour from that, although it's not even being considered. Rclone options come to mind and others. So I think it's valuable to warn about that.

Maybe something like: https://crates.io/crates/serde_path_to_error can be helpful here.

aawsome: I agree we should just keep the feature flag.

About the warning: In principle I agree about general options, but here we have options specific for a command which cannot even be called when not compiled with the feature...
I wonder how we should be able to print a warning. Imagine a user running `rustic backup` on a version without the `mount` feature. Should we warn "you have [mount] in your profile which is not used"? I mean when users call `rustic mount` with that build, they will already get an error and `[mount]` is only ever used in the `mount` command...
And users using a version with the `mount` feature would call `rustic backup` without getting a warning about `[mount]` in the profile - even though their called command does not use any of the options set there...?!?

I have now added a `Value` field in the case without the feature - that will suppress the error but also don't print any warning...

aawsome:
rebased on top of #1329

src/commands/mount.rs
simonsan: I'm thinking more of a `println` than a warning I think, as an `INFO`. But independent of the logging. So users become aware of that and can remove it manually from the profile if they want.

simonsan: That being said, this is just the way I think it makes sense to do it, for informing the user, about values in the config, that have no effect. We don't need to do it that way. Also, we may need to refactor some things, regarding the config parsing step for this, to be able to spit out better errors/warnings? So probably nothing for this PR here.

src/commands/mount.rs
```diff
@@ -0,0 +1,148 @@
...
+    snapshot_path: Option<String>,
+}
+
+impl Override<RusticConfig> for MountCmd {
+    // Process the given command line options, overriding settings from
+    // a configuration file using explicit flags taken from command-line
+    // arguments.
+    fn override_config(&self, mut config: RusticConfig) -> Result<RusticConfig, FrameworkError> {
+        let mut self_config = self.clone();
+        // merge "webdav" section from config file, if given
```
simonsan, line 60:
```suggestion
        // merge "mount" section from config file, if given
```

simonsan:
Some things that I think we should do to prepare for this PR:

- [ ] add build support in CI/CD 
  - [ ] add to be installed dependencies to OS build matrix
  - [ ] add feature flags to OS build matrix

tbc

simonsan (approved):

src/commands/mount.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    /// How to handle access to files. [default: "forbidden" for hot/cold repositories, else "read"]
+    #[clap(long)]
+    #[merge(strategy=conflate::option::overwrite_none)]
+    file_access: Option<String>,
+
+    /// The mount point to use
+    #[clap(value_name = "PATH")]
+    #[merge(strategy=conflate::option::overwrite_none)]
+    mountpoint: Option<PathBuf>,
```
simonsan, line 46: `mount_point`

PR #985, opened 2024-01-03
Correct typo

kapitainsky:
Correct typo


kapitainsky:
OK. Thx.

I am new to rustic and when tried to install it using `binstall` I had this error :

```
$ cargo binstall rustic-rs
error: no such command: `binstall`

	Did you mean `install`?

	View all installed commands with `cargo --list`
```

so I assumed that it must be typo:)

PR #1019, opened 2024-01-26
feat: Add option --numeric-uid-gid to ls

aawsome:
see https://github.com/rustic-rs/rustic/discussions/503#discussioncomment-8254371


aawsome (commented):
LGTM

Issue #859, opened 2023-08-30
Add `RUSTIC_HOME` and update precedence for platform specific app directories


simonsan:
> it gives the possibility to set a location of their own to the users.

yeah, users can set a custom location with the environment variable, but `"RUSTIC_HOME": "$dir"` is defined in manifest,  `$env:RUSTIC_HOME` will be reset to `$dir` every update.

---

If you want to make the program more portable. `rustic_home` should be set in order and fallback to the first existing path:

0. `$env:RUSTIC_HOME`
1. `<program_root>\home`
2. `$env:USERPROFILE\.config\rustic` or `$env:USERPROFILE\.rustic`
3. `$env:APPDATA\rustic`
4. `$env:ProgramData\rustic`
...

note, this just a example or suggestion, or a feature request for your project.

_Originally posted by @HUMORCE in https://github.com/ScoopInstaller/Main/pull/5042#discussion_r1310120538_

Furthermore, we need to make sure, that the next release:
- [x] actually contains the hashes
- [x] has the platform specific binary extension (for Windows) added (should be fixed in #843)

tech189:
Hi, just to let you know, the latest release (0.6.1) doesn't contain any binaries for Windows so the Scoop package is out of date currently.

simonsan:
> yeah, users can set a custom location with the environment variable, but "RUSTIC_HOME": "$dir" is defined in manifest, $env:RUSTIC_HOME will be reset to $dir every update.

We need to think that through, as that could become a problem. We may only set `RUSTIC_HOME` in scoop if it's unset. On the other hand, it could also lead to issues, to not reset it on an update, as people could copy their executable from their scoop directory to use it from another folder. After updating, they would think, they are using the updated version — but they aren't. Which could lead to weird bug reports down the line. 

@tech189 Is there a usual way to handle that for `scoop` manifests?

tech189:
Installations of apps with Scoop are not meant to be portable as far as I know. Users should know that if they move executables away from their Scoop directory, they will no longer be managed by Scoop and updates/environment variables will no longer work properly.

magrathean-uk:
I opened a focused draft PR for this shared config-path fix: https://github.com/rustic-rs/rustic/pull/1867

It covers the RUSTIC_HOME/XDG precedence path and the Windows USERPROFILE/.config fallback.

PR #1059, opened 2024-02-11
feat(env): Make rustic more portable

simonsan:
To make rustic more portable, we can use `RUSTIC_HOME` to locate the user's wanted config directories without relying on conventional/standard directories varying on the operating system. On Windows, we add another layer to get compatibility paths (e.g. in the `USERPROFILE`).

Fixes https://github.com/rustic-rs/rustic/issues/859


src/config.rs
```diff
@@ -198,7 +208,42 @@ fn get_config_paths(filename: &str) -> Vec<PathBuf> {
             p
         })
     })
-    .collect()
+    .collect::<Vec<_>>();
+
+    paths.extend(dirs);
+    paths
```
simonsan, line 214: It could be also better here to extend `dirs` with `path`, instead of the other way around. I'd like for `RUSTIC_HOME` to be the top most.

aawsome: Yes, I think is best to create a vec of possible paths which are already ordered and then `filter_map` and `collect` as already done.

About the order, I agree: first the `RUSTIC_HOME` dir - and the additional windows dirs I actually would but last...

simonsan: Fixed (I think?)

aawsome: I think you should `extend` before `filter_map`...

simonsan: It's different types, win compatibility is already `Vec<PathBuf>` whereas the other paths are `Vec<Option<PathBuf>>`, hence the filter_map.

simonsan:
```rust
#[cfg(target_os = "windows")]
fn get_windows_portability_config_directories() -> Option<Vec<PathBuf>> {
    std::env::var_os("USERPROFILE").map(|path| {
        vec![
            PathBuf::from(path.clone()).join(r".config\rustic"),
            PathBuf::from(path).join(".rustic"),
        ]
    })
}
```

aawsome: The cruical part is about the `map` which adds the profile filename to the path.
Maybe just change the function such that it returns `Vec<Option<PathBuf>>`?

src/config.rs
```diff
@@ -185,20 +185,58 @@ fn extend(left: &mut HashMap<String, String>, right: HashMap<String, String>) {
 ///
 /// A vector of [`PathBuf`]s to the config files
 fn get_config_paths(filename: &str) -> Vec<PathBuf> {
-    [
+    #[allow(unused_mut)]
+    let mut paths = vec![
```
simonsan, line 189:
```suggestion
    // we need this mut here, because we want to add data to the Vec depending on the OS
    // but our CI, running on Unix, doesn't see the extending the Vector
    #[allow(unused_mut)]
    let mut paths = vec![
```

src/config.rs
```diff
@@ -185,20 +185,61 @@ fn extend(left: &mut HashMap<String, String>, right: HashMap<String, String>) {
 ///
 /// A vector of [`PathBuf`]s to the config files
 fn get_config_paths(filename: &str) -> Vec<PathBuf> {
-    [
+    // we need this mut here, because we want to add data to the Vec
+    // depending on the OS. Our CI is running on Unix, and doesn't have knowledge of
+    // us mutating that vector
+    #[allow(unused_mut)]
+    let mut paths = vec![
+        get_home_config_path(),
```
aawsome, line 193: Why not simply add two entries
```
std::env::var_os("USERPROFILE").map(|path| {
PathBuf::from(path).join(r".config\rustic"))
std::env::var_os("USERPROFILE").map(|path| {
PathBuf::from(path).join(r".config\rustic"))
std::env::var_os("USERPROFILE").map(|path| {
PathBuf::from(path).join(r".config\rustic")),
std::env::var_os("USERPROFILE").map(|path| {
PathBuf::from(path).join(r".rustic"))
```
here? If this seems to heavy, create 2 functions which return these results.

simonsan: Sorry, I don't understand. Instead of what should we do that, you marked ` get_home_config_path(),`?

simonsan: Do you mean like that:

```rust
    let mut paths = vec![
        get_home_config_path(),
        ProjectDirs::from("", "", "rustic")
            .map(|project_dirs| project_dirs.config_dir().to_path_buf()),
        get_global_config_path(),
        Some(PathBuf::from(".")),
        #[cfg(target_os = "windows")]
        std::env::var_os("USERPROFILE").map(|path| PathBuf::from(path).join(r".config\rustic")),
        #[cfg(target_os = "windows")]
        std::env::var_os("USERPROFILE").map(|path| PathBuf::from(path).join(r".rustic")),
    ];
```

simonsan: If that was the idea, I found it a bit cleaner, although we need to have all that lint deactivated and mutable variable, to have that in its own function. Just because, we can have access to that, to just get the config paths singly, when we need them.

For example, this:

```rust
fn get_config_paths(filename: &str) -> Vec<PathBuf> {
    vec![
        std::env::var_os("RUSTIC_HOME").map(|home_dir| PathBuf::from(home_dir).join("config")),
        ProjectDirs::from("", "", "rustic")
            .map(|project_dirs| project_dirs.config_dir().to_path_buf()),
        get_global_config_path(),
        Some(PathBuf::from(".")),
        #[cfg(target_os = "windows")]
        std::env::var_os("USERPROFILE").map(|path| PathBuf::from(path).join(r".config\rustic")),
        #[cfg(target_os = "windows")]
        std::env::var_os("USERPROFILE").map(|path| PathBuf::from(path).join(".rustic")),
    ]
    .into_iter()
    .filter_map(|path| path.map(|p| p.join(filename)))
    .collect::<Vec<_>>()
}
```

would work, but would not let us have `RUSTIC_HOME` access in a function, if we just need it for something else, e.g. opening a pdf in that folder, when we would want to open documentation or anything else in the home folder.

Hence I think extracting these things into their own functions is better long-term.

aawsome: Yes, then simply use `get_userprofile_config_rustic()` and `get_userprofile_rustic()` or something like this.

simonsan: Thinking about it, all those directories are a bit different:

`RUSTIC_HOME`: we use `RUSTIC_HOME/config` here for the configs
`USERPROFILE\.config\rustic`: we use a root directory here for the configs
`USERPROFILE\.rustic`: same here, we use a root directory

I think consistency is important in that regard, so people can for example use their `rustic-package` unpack it, change some configs and start. Which would mean, that we would need `config/` as a subdirectory in each of these, although it doesn't make a lot of sense in 
`USERPROFILE\.config\rustic` to have another subdir for `config`. But then, we also can't use this directory as a kind of HOME directory, where for example also `log` files would be found and other resources. 🤔

simonsan:
`Application::Path` with `RusticPaths` seems like the cleaner solution, so we have search paths easily accessible. Closing this here in favour of #1342

PR #1060, opened 2024-02-11
test: refactor integration tests to assert_cmd and predicates, test all configs in config subdirectory

simonsan:
The intention of this PR is to make it easier (more straight forward) to write `rustic` integration tests by utilizing the `assert_cmd` and `predicates` crates.


tests/backup_restore.rs
```diff
@@ -7,49 +7,39 @@
...
-    assert_eq!(
-        matches,
-        vec![(PatternID::must(0), 19), (PatternID::must(1), 21),]
-    );
+    rustic_runner(&temp_dir)?
+        .args(["init"])
+        .assert()
+        .success()
+        .stderr(predicate::str::contains("successfully created."))
+        .stderr(predicate::str::contains("successfully added."));
```
simonsan, line 41: Is that intentional, that we log that on stderr on init?

aawsome: Yes, it is all an `info!` log here. And logs are by default printed to stderr. You could even turn this message off by using `--log-level=warn/error/off`.

aawsome (approved):
LGTM! Thanks @simonsan

PR #1061, opened 2024-02-12
fix(commands): Ask for password in backup and copy command if it is missing

aawsome:
A user discovered that for some commands they are not being asked for a password, although they should. This implements a fix for the backup and copy command so they consistently call the same functionality that is being used for the other commands. Hence, asking for the password, if it's missing.

### DIFF
--- a/src/commands/backup.rs
+++ b/src/commands/backup.rs
@@ -3,7 +3,7 @@
 use std::path::PathBuf;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


simonsan (approved):

aawsome:
Note that `rustic copy --init` still doesn't ask for the destination password when initializing the destination repository (and it is not specifed).
To implement this, we need more rustic_core support...

PR #1072, opened 2024-02-19
fix(doc): Correct b2.toml

### LINKED ISSUE #1068: Example b2.toml issues
https://github.com/rustic-rs/rustic/blob/main/config/services/b2.toml

What's the difference between `application_key` and `application_key_id`?

Both are required:

    error: backend openDAL Backend cannot be loaded: ConfigInvalid (permanent) at Builder::build, context: { service: b2 } => application_key_id is empty

    error: backend openDAL Backend cannot be loaded: ConfigInvalid (permanent) at Builder::build, context: { service: b2 } => application_key is empty

However, `application_key_id` is not provided in the example.

`".my_key"` is given as an example value for the `application_key` field.  Is the intent that the value of this key is a filename?  

If so, what is the directory name that `.my_key` is relative to?

HaleTom:
closes #1068

### DIFF
--- a/config/services/b2.toml
+++ b/config/services/b2.toml
@@ -3,7 +3,7 @@
 repository = "opendal:b2" # just specify the opendal service here
 password = "<rustic_passwd>"
 # or 
-# password-file = "/home/<username>/etc/secure/restic_passwd"
+# password-file = "/home/<username>/etc/secure/rustic_passwd"
 
 # B2 specific options
 [repository.options]
--- a/config/services/b2.toml
+++ b/config/services/b2.toml
@@ -8,9 +8,6 @@ password = "<rustic_passwd>"
 # B2 specific options
 [repository.options]
 # Here, we give the required b2 options, see https://opendal.apache.org/docs/rust/opendal/services/struct.B2.html
-# These are required either here or set by environment variables
-# - See: https://github.com/apache/opendal/blob/39e0a69c8eadc44391ba0420fdc7de2a42962ecf/.env.example#L173-L178
-# - Issue pending: https://github.com/rustic-rs/rustic/issues/1073
 application_key_id = "my_id"  # B2 application key ID
 application_key = "my_key"    # B2 application key secret
 bucket = "bucket_name"        # B2 bucket name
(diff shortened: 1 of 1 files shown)


HaleTom:
Related:

rustic: OPENDAL_B2_* variables don't work #1073
https://github.com/rustic-rs/rustic/issues/1073

opendal: Do OPENDAL_B2_* environment variables work? #4199
https://github.com/apache/opendal/issues/4199

config/services/b2.toml
```diff
@@ -1,14 +1,18 @@
-# rustic config file to use opendal storage - in this case b2
+# rustic config file to use B2 storage via Apache OpenDAL
 [repository]
 repository = "opendal:b2" # just specify the opendal service here
 password = "mypassword"
+# or 
+# password-file = "/home/<username>/etc/secure/restic_passwd"
```
aawsome, line 6: maybe name it `rustic_passwd`? People using restic for sure will know that they can also use the restic password file here.

config/services/b2.toml
```diff
@@ -1,14 +1,18 @@
...
 [repository.options]
 # Here, we give the required b2 options, see https://opendal.apache.org/docs/rust/opendal/services/struct.B2.html
-root = "repo-path" # Set the work directory for backend
-key_id = "my_id" # B2 application key keyID
-application_key = ".my_key" # B2 application key applicationKey
-bucket = "bucket_name" # B2 bucket name
-bucket_id = "bucket_id" # B2 bucket_id
+# These are required either here or set by environment variables
+# - See: https://github.com/apache/opendal/blob/39e0a69c8eadc44391ba0420fdc7de2a42962ecf/.env.example#L173-L178
+# - Issue pending: https://github.com/rustic-rs/rustic/issues/1073
```
aawsome, line 13: Please remove it here. Once we takle #1073 we can add it. Current state is it isn't available through env variables.

aawsome (changes requested):
Thanks @HaleTom for the PR! I have only minor change requests.

aawsome (approved):
Thanks a lot @HaleTom

PR #1078, opened 2024-02-22
feat(commands): Add --check-index option

aawsome:
Adds the `--check-index` option. This is an alternative to `rustic repair index` in cases where the repository is read-only and cannot be repaired. 

Using this option, the index is checked against existing pack files and missing information is read from the pack headers.


aawsome (commented):
LGTM

Issue #1093, opened 2024-03-07
Render TOML version of `show-config`


AlphaJack:
Hi, would it be possible to get the output of `show-config` in JSON format? Right now the `--json` flag is not supported

aawsome:
Actually I think it would be better to output some TOML - as the file format is toml, too. And I think it would be better to rename as it is a bit ambiguous (there is the profle config we are talking here and the in-repo config which can be set by `rustic config`). Maybe something like `show-profile`? or `profile show-config`?

simonsan:
I agree with the `toml` choice, also because you can output it via `>>` to another file from that to copy a config. With the renaming, I would wait until we think about the cli refactor (if we want to go down this route)?

aawsome:
Is implemented via #1095. I close this issue - the renaming of commands is another concern.

PR #1095, opened 2024-03-07
feat(commands)!: show-config now outputs toml

aawsome:
`show-config` now outputs toml.

see #1093


src/commands/show_config.rs
```diff
@@ -11,6 +12,6 @@ pub(crate) struct ShowConfigCmd {}
 impl Runnable for ShowConfigCmd {
     fn run(&self) {
         let config = RUSTIC_APP.config();
-        println!("{config:#?}");
+        println!("{}", to_string_pretty(config.as_ref()).unwrap());
```
simonsan, line 15: What about something like this? Or did you unwrap intentionally because of `human-panic`?

```suggestion
        let Ok(config) = to_string_pretty(config.as_ref()) else {
            status_err!("An error occured, config cannot be shown.");
            PACE_APP.shutdown(Shutdown::Crash);
        }
        
        println!("{config}");
```

simonsan: Or go via `match`:


```suggestion
        match to_string_pretty(config.as_ref()) {
            Ok(config) => println!("{config}"),
            Err(err) => {
                status_err!("An error occured, config cannot be shown: {err}");
                PACE_APP.shutdown(Shutdown::Crash);
            },
        }
        

```

simonsan: Another route:

```rust
impl Display for RusticConfig {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        let Ok(config) = toml::to_string_pretty(self) else {
            return write!(f, "Error: Could not serialize config to TOML");
        };

        write!(f, "{}", config)
    }
}
```

simonsan: I wrote it on Github as a suggestion, so didn't run any checks or formatting

src/filtering.rs
```diff
@@ -87,7 +87,10 @@ impl SnapshotFilter {
     #[must_use]
     pub fn matches(&self, snapshot: &SnapshotFile) -> bool {
         if let Some(filter_fn) = &self.filter_fn {
-            match filter_fn.call::<bool>(snapshot) {
+            match SnapshotFn::from_str(filter_fn)
```
aawsome, line 90: Actually this is quite expensive, so it should be somehow preprocessed...

src/filtering.rs
```diff
@@ -87,7 +87,10 @@ impl SnapshotFilter {
...
+                .unwrap()
```
aawsome, line 91: Also, error handling should be not by panicking...

PR #1106, opened 2024-03-15
fix: Create dirs for log-file

### LINKED ISSUE #1105: `The system cannot find the path specified` is not providing info about the problematic path
Hi, I'm trying to launch rustic on windows, but I receive the error below:

```
rustic --log-level debug --use-profile "D:/Configurations/Test" show-config
error: rustic-rs fatal error: I/O operation failed: The system cannot find the path specified. (os error 3)
```

Is it possible to have the problematic path printed, at least with `debug` on, as it could be either the source or the destination?

I know it does not refer to the profile path, otherwise it would complain with:

```
rustic --log-level debug --use-profile "D:/Configurations/non-existent" show-config
[WARN] using no config file, none of these exist: D:/Configurations/non-existent.toml, D:/Configurations/rusticfds.toml, D:/Configurations/non-existent.toml
```

A minor issue is that it prints the same path 3 times if an absolute profile path is given

aawsome:
rustic now creates needed dirs for the logfile

closes #1105

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -25,7 +25,7 @@ pub(crate) mod tag;
 #[cfg(feature = "webdav")]
 pub(crate) mod webdav;
 
-use std::fs::File;
+use std::fs::{create_dir_all, File};
 use std::path::PathBuf;
 use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


aawsome:
Closing this as #1111 is the better solution.

simonsan:
> Closing this as #1111 is the better solution.

Agreed!

PR #1114, opened 2024-04-06
feat: Add interactive snapshots mode

aawsome:
Adds an interactive (TUI) mode. E.g. call `rustic snapshots -i`

Currently only a snapshot view is implemented with possibilities to alter snapshots (label, tags, delete protection mark).

Use `--filter-*` to filter snapshots in the interactive mode


aawsome:
Example screenshot: 
image

aawsome (commented):
LGTM

aawsome:
I'm merging this as it is a very independent (and new) part. There might be bugs and open improvements, but IMO it is already a good starting point.

PR #1124, opened 2024-04-18
fix(config)!: multiple paths in config profile as array

### LINKED ISSUE #1094: error sanitizing source with whitespaces in config file
#558 

I have a folder with whitespace that I would like to backup:

```toml
[[backup.sources]]
source = "/home/jack/Mobile Backup"
```

CLI output:
```
[00:00:00] reading index...               ████████████████████████████████████████          0/0                                                       
[WARN] error sanitizing source="/home/jack/Mobile Backup" in config file
```

Escaping with a backslash works only if I use single quotes:

```toml
[[backup.sources]]
source = '/home/jack/Mobile\ Backup'
```
CLI output:
```
[00:00:00] reading index...               ████████████████████████████████████████          0/0                                                       
[00:00:00] getting latest snapshot...     ████████████████████████████████████████          0/0                                                       
[00:00:17] backing up...
```

But it doesn't work with double quotes:

```toml
[[backup.sources]]
source = "/home/jack/Mobile\ Backup"
```
CLI output:
```
error: rustic-rs fatal error: parse error: invalid escape character in string: ` ` at line 8 column 46
```

In the [full configuration file example](https://github.com/rustic-rs/rustic/blob/6fffaad1cc1aeebf874063fc62125f4ce3b490b5/config/full.toml#L115C1-L115C93) it is said that `source` can contain multiple path in a string, separated by a whitespace, and I think this is the cause of the issue.

Why not having `sources` as an array of `source`, as `password-command` is an alternative to `password`?

### LINKED ISSUE #1122: "Permission denied" for source with many paths (including spaces)
[This config](https://gist.github.com/NatureHog/c4073c64101c687e723e0eb9a55a78b5) always gives a "Permission denied" error, no matter what I try. So far I've tried running as sudo, changing the backend (using rclone normally), changing the sources from "~" to "/home/USERNAME" and back.

The weird thing is, using the normal CLI works just fine, even using the same backup sources.

Using Linux Mint 21.3, rustic 0.7.0, backend rclone

I'll be happy to provide more details, if needs be.

Thanks in advance!

aawsome:
-- BREAKING CHANG --

Using multiple paths for a sources in the config file caused many problems as the syntax was unclear and problems existed in edge cases.

This PR changes the definition of sources in config profile files:
- a single path path can be specified as before, i.e. using `source = "/my/path"`.
- multiple paths must now be given in an array:
```
source = ["/my/path1", "/my/path2"]
```

Note that 
```
source = "/my/path1 /my/path2"
```
is now interpreted a a single path using the dir-tree "my","path1 " (with space), "my", "path2".

closes #1122
closes #1094

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -152,9 +152,9 @@ can be overwritten in the source-specifc configuration, see below.
 **Note**: All of the backup options mentioned before can also be used as
 source-specific option and then only apply to this source.
 
-| Attribute | Description                          | Default Value | Example Value         |
-| --------- | ------------------------------------ | ------------- | --------------------- |
-| source    | Source directory or file to back up. | Not set       | "/tmp/dir/to_backup/" |
+| Attribute | Description                          | Default Value | Example Value               |
+| --------- | ------------------------------------ | ------------- | --------------------------- |
+| source    | Source directory or file to back up. | Not set       | "/dir" , ["/dir1", "/dir2"] |
 
 ### Forget Options
 
--- a/config/full.toml
+++ b/config/full.toml
@@ -113,7 +113,10 @@ label = "label" # Default: not set
 # .. and so on. see [backup]
 
 [[backup.sources]]
-source = "/path/to/source2 /second/path" # multiple local paths are allowd within one source
+source = [
+  "/path/to/source2",
+  "/second/path",
+] # multiple local paths are given as array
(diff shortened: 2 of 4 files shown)


aawsome (commented):
LGTM

PR #1130, opened 2024-04-22
feat!: Allow specifying many options in config profile without array

aawsome:


aawsome (commented):
LGT

intgr:
> password-command and warmup-command now must be given as array if the commands use args.

To be honest, I think this is slightly annoying. I can no longer copy-paste the command directly into a shell to test that it works.

aawsome:
> To be honest, I think this is slightly annoying. I can no longer copy-paste the command directly into a shell to test that it works.

@intgr I got your point. So maybe we should allow both: a single string which would be automatically split into its parts and an array to cover complicated spacing.... I'll discuss this with the other maintainers..

aawsome:
@intgr After a short discussion we decided that we want to support a single string. So this change will be kind of reverted/overworked.

thePanz:
Thanks for working in supporting the single string configuration for "password-command" again.

Would be great to also include a more clear message when the failure related to that configuration change is triggered.

Example: I still had the "old" configuration with `password-command = "secret-tool lookup ..."`, which now it gives just a single line of error like when running `rustic`:

`error: No such file or directory (os error 2)`

It took a while to identify where the error was triggered, and the "no such file" was quite misleading :exploding_head:
Adding some context on the error might help

intgr:
> It took a while to identify where the error was triggered, and the "no such file" was quite misleading 🤯

I also experienced this and submitted a PR to improve the situation: https://github.com/rustic-rs/rustic_core/pull/265

PR #1131, opened 2024-04-22
feat!(command): copy: Use config profile as target

aawsome:
Breaking change:

Targets for the `copy` command must now be given by using config profile(s).

### DIFF
--- a/src/commands/copy.rs
+++ b/src/commands/copy.rs
@@ -2,40 +2,58 @@
 
 use crate::{
     commands::{get_repository, init::init_password, open_repository, open_repository_indexed},
-    config::AllRepositoryOptions,
     helpers::table_with_titles,
-    status_err, Application, RUSTIC_APP,
+    status_err, Application, RusticConfig, RUSTIC_APP,
 };
-use abscissa_core::{Command, Runnable, Shutdown};
+use abscissa_core::{config::Override, Command, FrameworkError, Runnable, Shutdown};
 use anyhow::{bail, Result};
-use log::{error, info};
+use log::{error, info, log, Level};
 use merge::Merge;
 use serde::{Deserialize, Serialize};
+use serde_with::{serde_as, OneOrMany};
 
 use rustic_core::{CopySnapshot, Id, KeyOptions};
 
 /// `copy` subcommand
-#[derive(clap::Parser, Command, Debug)]
-pub(crate) struct CopyCmd {
+#[serde_as]
+#[derive(clap::Parser, Command, Default, Clone, Debug, Serialize, Deserialize, Merge)]
+pub struct CopyCmd {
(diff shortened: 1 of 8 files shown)


src/commands/copy.rs
```diff
@@ -51,8 +69,8 @@ impl CopyCmd {
     fn inner_run(&self) -> Result<()> {
         let config = RUSTIC_APP.config();
 
-        if config.copy.targets.is_empty() {
-            status_err!("no [[copy.targets]] section in config file found!");
+        if config.copy.target.is_empty() {
+            status_err!("please specify at least 1 target!");
```
simonsan, line 73: I think that error message should make the actual cause of the error more explicit.
Something along the lines:

"Copy target must be a standalone rustic profile, target {target} doesn't exist."

aawsome: Here, `target` is just a `String`, but I adapted the error message to be more clear.

Also, when the profile cannot be loaded, now an `error` log is displayed:
```
[INFO] using config /home/alex-dev/.config/rustic/default.toml
[INFO] repository local:/tmp/repo: password is correct.
[INFO] using cache at /home/alex-dev/.cache/rustic/cc2abcdea25343bcc6072cbeb3452fb6c735aea7682d1653a10b01e980703e6c
[00:00:00] reading index...               ████████████████████████████████████████          2/2                                                                             
[00:00:00] getting snapshots...           ████████████████████████████████████████          4/4                                                                             
[ERROR] using no config file, none of these exist: /home/alex-dev/.config/rustic/xxx:.toml, /etc/rustic/xxx:.toml, ./xxx:.toml
error: No repository given.
```

tests/show-config-fixtures/empty.txt
```diff
@@ -51,7 +51,7 @@ sources = []
 source = []
 
 [copy]
-targets = []
+target = []
```
simonsan, line 54: I think that should be still plural, as we have an array of values here. Although the array itself might make it clear, I think it's clearer to actually name it `targets` as well.

aawsome: Having singular or plural is always a tough decission... I agree that it can be targets here.
But `target = "target_profile"` also works (due to `serde_as::OneOrMany`). And the CLI also has this as option `--target` (description: `Target repository (can be specified multiple times)`).

So, the question is what is more strange:
- `[copy] targets = "my_target"`
- `[copy] target = ["my_target"`, "my_target2"]
- `rustic copy --targets=my_target --targets=my_target2`

Moreover I'd like to be consistent with other options where we basically get arrays but serde/clap also allows to handle single results.

Currently, we use singular in most places, IIRC.

aawsome: is changed to `targets` and `--target` in clap.

simonsan (approved):

PR #1137, opened 2024-04-25
feat(interactive): Add possibility to change snapshot description

### LINKED ISSUE #1134: Add Description to Existing Snapshot
Hi,

From what I can tell, the `--description` option is only available when creating a new snapshot. As I would prefer to use **resticprofile** for taking the actual snapshot, is there any way to add a description to an existing snapshot? Thanks

aawsome:
closes #1134


aawsome (commented):
LGTM

PR #1139, opened 2024-04-26
chore: Remove self-update from default crate features

alerque:
The self-updater feature is problematic for distro package builders where the executable does not have permission and should not be trying to monkey with the system installed binaries. Distros have their own update mechanisms that shouldn't be tampered with by every app that comes along. This means distro have to build with --no-default-features for apps that include self-updaters by default, but that also means we have to maintain a list of features we do want. This is not only tedious it is error-prone because there is a very good chance of new features getting overlooked when doing version bumps.

Setting up a feature group like this makes it much easier for distros to build without the unwanted features without getting out of sync with upstream enhancements over time.


alerque:
Personally I would actually recommend taking the self updater *out* of the default feature list entirely. Folks can and should get their updates by whatever means they first used to get it (distro packages, `cargo install`, source builds, etc.). The only exception I would make is for the upstream built binaries attached to releases. Those probably can and should have the feature enabled. That means adding the feature manually to the CI builds, but setting it to be off by default in everything else. I made this PR to be the least intrusive change possible, but if you would accept the change to the default feature list I would be happy to adapt it for this recommendation.

simonsan:
Personally, I'm up for adapting default features in the way, that you can package with default-features instead of using another newly introduced feature. I'm myself not a very big fan of having self-updatable executables. So I understand where you are coming from. Let's wait for @aawsome opinion though, and if he agrees we can use the mentioned changes to the default features, instead.

I can update our CI/CD workflows then.

alerque:
Thanks for the review and being willing to re-consider the defaults. I'll update my changes with that in mind.

alerque:
I've replaced 5c1089d with 8be7b74 which just drops the self-updater from the default flag list.

I had a look at CI but it isn't apparent to me what ends up where. None of the RPM or other *package* builds should have this feature enabled (as the meta data for the package would be out of sync even if somehow the binary got updated). On the other hand any binaries that the end use would download and place manually themselves are candidates for having this re-enabled.

@simonsan Feel free to tag team on this PR or point me in the right direction or whatever...

aawsome (approved):
LGTM

simonsan:
This needs to wait a bit with merging though, as I need to update the CD first to build rustic-nightly and release with the corresponding added features. Will merge it when ready.

PR #1175, opened 2024-07-09
fix(tui): reset terminal no matter what

### LINKED ISSUE #1166: Interactive Snapshots without Specifying Repository Bug
I am using the nightly build, just upgraded to the latest release today. When I tried to use `rustic snapshots -i` without specifying a profile or repository, my terminal went haywire. It's not a critical issue as long as I remember to include a repository but thought you should be aware. Not sure how to describe it so here are some screenshots:

After entering `rustic snapshots -i`:

image

The terminal does not show any input and when I pressed enter on the above screen:

image

Then reconnecting to the session:

image

drdo:
Currently if you run `rustic snapshots -i` without a repository configured `rustic` will leave the terminal in raw mode.

This is a simple fix to make sure the terminal is properly reset in all situations.

close #1166


aawsome (approved):
Thanks @drdo!
LGTM

PR #1201, opened 2024-08-18
chore: update rustic_core and rustic_backend

aawsome:


tests/show-config-fixtures/empty.txt
```diff
@@ -59,24 +59,6 @@ filter-host = []
...
-keep-half-yearly = 0
-keep-yearly = 0
-keep-within = "0s"
-keep-within-hourly = "0s"
-keep-within-daily = "0s"
-keep-within-weekly = "0s"
-keep-within-monthly = "0s"
-keep-within-quarter-yearly = "0s"
-keep-within-half-yearly = "0s"
-keep-within-yearly = "0s"
```
nardoor, line 79: Could you please give some context as for why you needed to remove these? I am wondering ^^

Cargo.toml
```diff
@@ -41,8 +41,8 @@ rustdoc-args = ["--document-private-items", "--generate-link-to-definition"]
 
 [dependencies]
 abscissa_core = { version = "0.7.0", default-features = false, features = ["application"] }
-rustic_backend = { git = "https://github.com/rustic-rs/rustic_core.git", features = ["cli"] }
-rustic_core = { git = "https://github.com/rustic-rs/rustic_core.git", features = ["cli"] }
```
nardoor, line 45: This is a good change I think.
Earlier some changes in `core` or `backend` could break  `rustic-rs` without even changing the code here.

nardoor (approved):

tests/show-config-fixtures/empty.txt
aawsome: I changed the `KeepOptions` in rustic_core in https://github.com/rustic-rs/rustic_core/pull/238. There also the Serialization was changed to skip them if the Options are `None`. The serialization is actually only relevant for the `show-config` command in rustic (and for the unit tests added in that PR).
For `show-config` we do have integration tests here in rustic, so I needed to change the expected output - after updating to the rustic_core version which includes that PR.

Cargo.toml
aawsome: Yes. We do it that during development versions, rustic uses git dependencies on rustic_core and rustic_backend. This is because we dont't want to release each single improvement in those crates in order to have them available for rustic development versions.

However, when releasing rustic, we need version dependencies because git dependencies are not allowed by crates.io.

nardoor: Oh yeah so it's planned to revert to git dependencies after the release I see.

tests/show-config-fixtures/empty.txt
nardoor: Thanks I understand better!

PR #1216, opened 2024-08-29
feat(interactive): Allow to view text files

aawsome:
Adds the possibility to view text files up to 1MiB  in the interactive snapshot tree view. Note this is a quick-and-dirty implementation which could be improved better non-text-files and which could be optimized to read files of all sizes

### DIFF
--- a/src/commands/tui/ls.rs
+++ b/src/commands/tui/ls.rs
@@ -41,7 +41,7 @@ General Commands:
       q,Esc : exit
       Enter : enter dir
   Backspace : return to parent dir
-          v : view file contents (up to 1MiB)
+          v : view file contents (text files only, up to 1MiB)
           r : restore selected item
           n : toggle numeric IDs
           ? : show this help page
@@ -217,18 +217,20 @@ impl<'a, P: ProgressBars, S: IndexedFull> Snapshot<'a, P, S> {
                                     if let Ok(data) = OpenFile::from_node(self.repo, node).read_at(
                                         self.repo,
                                         0,
-                                        node.meta.size.min(1000000).try_into().unwrap(),
+                                        node.meta.size.min(1_000_000).try_into().unwrap(),
                                     ) {
-                                        let content = String::from_utf8_lossy(&data);
-                                        let lines = content.lines().count();
-                                        let path = self.path.join(node.name());
-                                        let path = path.display();
-                                        self.current_screen =
-                                            CurrentScreen::ShowFile(popup_scrollable_text(
-                                                format!("{}:/{path}", self.snapshot.id),
-                                                &content,
-                                                (lines + 1).min(40).try_into().unwrap(),
-                                            ));
+                                        // viewing is only supported for text files
+                                        if let Ok(content) = String::from_utf8(data.to_vec()) {
+                                            let lines = content.lines().count();
+                                            let path = self.path.join(node.name());
+                                            let path = path.display();
+                                            self.current_screen =
+                                                CurrentScreen::ShowFile(popup_scrollable_text(
+                                                    format!("{}:/{path}", self.snapshot.id),
+                                                    &content,
+                                                    (lines + 1).min(40).try_into().unwrap(),
+                                                ));
+                                        }
                                     }
                                 }
                             }
--- a/src/commands/tui/ls.rs
+++ b/src/commands/tui/ls.rs
@@ -5,6 +5,7 @@ use crossterm::event::{Event, KeyCode, KeyEventKind};
 use ratatui::{prelude::*, widgets::*};
 use rustic_core::{
     repofile::{Node, SnapshotFile, Tree},
+    vfs::OpenFile,
     IndexedFull, ProgressBars, Repository,
 };
 use style::palette::tailwind;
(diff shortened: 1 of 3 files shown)


src/commands/tui/widgets/text_input.rs
```diff
@@ -40,22 +50,29 @@ impl ProcessEvent for TextInput {
...
+                (Enter, _) if self.lines == 1 => {
                     return TextInputResult::Input(self.textarea.lines().join("\n"));
                 }
-                KeyEvent {
-                    code: Char('s'),
-                    modifiers: KeyModifiers::CONTROL,
-                    ..
-                } => {
+                (Char('s'), KeyModifiers::CONTROL) => {
                     return TextInputResult::Input(self.textarea.lines().join("\n"));
```
nardoor, line 64: for an input that isn't `changeable`, `Ctrl + s` will close the the `TextInput` popup. Which could be a bit misleading?
I'd advise that nothing should be done for an immutable popup when `Ctrl + s`

What do you think ?

src/commands/tui/widgets/text_input.rs
```diff
@@ -40,22 +50,29 @@ impl ProcessEvent for TextInput {
     type Result = TextInputResult;
     fn input(&mut self, event: Event) -> TextInputResult {
         if let Event::Key(key) = event {
+            let KeyEvent {
+                code, modifiers, ..
+            } = key;
             use KeyCode::*;
-            match key {
```
nardoor, line 44: the `changeable` parameter has heavy consequences on input interpretation.

For this reason, we could want to have completely separated `match (code, modifiers)`  for `changeable` and `!changeable` cases

src/commands/tui/widgets/text_input.rs
```diff
@@ -15,12 +16,21 @@ pub enum TextInputResult {
 }
 
 impl TextInput {
-    pub fn new(text: &str, initial: &str, lines: u16) -> Self {
+    pub fn new(text: Option<&str>, initial: &str, lines: u16, changeable: bool) -> Self {
```
nardoor, line 19: Design question: 

A `TextInput` that is not `Changeable` isn't really an `Input`.

Should `TextInput` be renamed in `TextPopup`? Or should there be two separated structures, implementing similar traits?

src/commands/tui/ls.rs
```diff
@@ -204,6 +209,31 @@ impl<'a, P: ProgressBars, S: IndexedFull> Snapshot<'a, P, S> {
...
                     Char('n') => self.toggle_numeric(),
+                    Char('v') => {
+                        // viewing is not supported on cold repositories
+                        if self.repo.config().is_hot != Some(true) {
+                            if let Some(node) = self.selected_node() {
+                                if node.is_file() {
+                                    if let Ok(data) = OpenFile::from_node(self.repo, node).read_at(
+                                        self.repo,
+                                        0,
+                                        node.meta.size.min(1000000).try_into().unwrap(),
```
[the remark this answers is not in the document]
aawsome: is changed.

src/commands/tui/widgets.rs
```diff
@@ -49,7 +49,18 @@ pub fn popup_input(
     lines: u16,
 ) -> PopUpInput {
     PopUp(WithBlock::new(
-        TextInput::new(text, initial, lines),
+        TextInput::new(Some(text), initial, lines, false),
```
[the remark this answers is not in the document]
aawsome: Actually, `false` was wrong here. Is changed to `true`.

src/commands/tui/widgets/text_input.rs
aawsome: is changed, see below.

src/commands/tui/widgets/text_input.rs
aawsome: Good idea! I made two blocks how to handle keys.
When we see that there are more differences to come-up, we can always split the widgets in future..

src/commands/tui/widgets/text_input.rs
aawsome: Good idea! I made two blocks how to handle keys.

src/commands/tui/ls.rs
```diff
@@ -204,6 +209,31 @@ impl<'a, P: ProgressBars, S: IndexedFull> Snapshot<'a, P, S> {
...
+                                    ) {
+                                        let content = String::from_utf8_lossy(&data);
```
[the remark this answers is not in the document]
aawsome: I changed it so it does open only for text files.

aawsome:
@nardoor Thanks for the review, especially for finding the bug about the `changeable` :-)

src/commands/tui/ls.rs
```diff
@@ -217,18 +217,20 @@ impl<'a, P: ProgressBars, S: IndexedFull> Snapshot<'a, P, S> {
...
-                                        let path = self.path.join(node.name());
-                                        let path = path.display();
-                                        self.current_screen =
-                                            CurrentScreen::ShowFile(popup_scrollable_text(
-                                                format!("{}:/{path}", self.snapshot.id),
-                                                &content,
-                                                (lines + 1).min(40).try_into().unwrap(),
-                                            ));
+                                        // viewing is only supported for text files
+                                        if let Ok(content) = String::from_utf8(data.to_vec()) {
```
nardoor, line 223: What do you think about a popup that warns that the `view` action failed because the file isn't full utf8?

nardoor (approved):
apart from a potential warning that the `view` action couldn't be finalized as the file isn't `utf8`, it lgtm.

src/commands/tui/ls.rs
aawsome: Agree - but IMO this can go with other future user improvements. It may be a bit confusing but is doing no harm...

PR #1220, opened 2024-09-02
fix: return exitcode

### LINKED ISSUE #927: Rustic returns exit code 0 when errors occur
For automation tools it would be useful if a non-zero exit code is returned when the repository is not initialized. In fact this is exactly what Restic suggests you do in their [scripting documentation](https://restic.readthedocs.io/en/stable/075_scripting.html#check-if-a-repository-is-already-initialized), but it doesn't work for Rustic:

```
/app # restic -r test cat config
Fatal: unable to open config file: stat test/config: no such file or directory
Is there a repository at the following location?
test
/app # echo $?
1
/app # rustic -r test cat config
using no config file, none of these exist: /root/.config/rustic/rustic.toml, /etc/rustic/rustic.toml, ./rustic.toml
error: No repository config file found. Is there a repo at local:test?
/app # echo $?
0
```

There may be more instances where exit codes are not properly implemented, but this is all I have tested as of now.

aawsome:
rustic now returns an error code for failing command.

Note: Some command do not fail yet fail in every cases user would expect is. Most notably the `check` ckommand producing `error` messages may still return error code 0.

closes #927

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -113,8 +113,8 @@ impl Application for RusticApp {
 impl RusticApp {
     /// Shut down this application gracefully, exiting with given exit code.
     fn shutdown_with_exitcode(&self, shutdown: Shutdown, exit_code: i32) -> ! {
-        let result = self.state().components().shutdown(self, shutdown) {
-        if let Err(e) = result
+        let result = self.state().components().shutdown(self, shutdown);
+        if let Err(e) = result {
             fatal_error(self, &e)
         }
 
--- a/src/application.rs
+++ b/src/application.rs
@@ -1,11 +1,11 @@
 //! Rustic Abscissa Application
-use std::env;
+use std::{env, process};
 
 use abscissa_core::{
-    application::{self, AppCell},
+    application::{self, fatal_error, AppCell},
     config::{self, CfgCell},
     terminal::component::Terminal,
-    Application, Component, FrameworkError, StandardPaths,
+    Application, Component, FrameworkError, Shutdown, StandardPaths,
 };
 
 use anyhow::Result;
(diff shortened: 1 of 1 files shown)


src/application.rs
```diff
@@ -99,4 +99,24 @@ impl Application for RusticApp {
...
+
+impl RusticApp {
+    /// Shut down this application gracefully, exiting with given exit code.
+    fn shutdown_with_exitcode(&self, shutdown: Shutdown, exit_code: i32) -> ! {
+        if let Err(e) = self.state().components().shutdown(self, shutdown) {
+            fatal_error(self, &e)
+        }
+
+        process::exit(exit_code);
+    }
```
nardoor, line 121: I wonder if there is a value of adding this method instead of in-lining this code into the Application::shutdown implementation.

aawsome: Yes, you are right. Currently only `Application::shutdown` is used. However, in future we plan to add more exit codes than just 0 or 1. 

E.g. a `rustic check` which found no error (caveat: if errors are found there will be no error code 1 unless https://github.com/rustic-rs/rustic_core/pull/224 has been merged) but only warning.

Or a `rustic backup` which had problems with some files, but could in general backup most files.

For this and because of https://github.com/iqlusioninc/abscissa/pull/865/files (which unfortunately is still not released), I added this method...

nardoor: I understand better!
IMO, the fact that we're waiting for this abscissa pull request to be released could be specified in a comment?

Whether or not you add the comment, I'm okay with resolving my comment/merging.

nardoor (approved):
Lgtm (see: https://github.com/rustic-rs/rustic/pull/1220#discussion_r1747756185)

PR #1225, opened 2024-09-05
feat: Add autocompletion hints

nardoor:
## Feature

**Add auto completion hints**

This helps some shells to propose better completions.
For instance on `zsh`, no completion were proposed when doing `rustic backup <...>[TAB]`

## Hints added
- `rustic backup <...>[TAB]`
- `rustic backup --as-path [TAB] <...>`
- `rustic backup --stdin-filename [TAB] <...>` This one is for comfort, as the compl might not help (or might), but my opinion is that it's better than nothing.
- `rustic diff <snapshot_id> <...>[TAB]`
- `rustic find --path <...>[TAB]`
- rustic --log-file <...>[TAB]`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -560,9 +560,9 @@ dependencies = [
 
 [[package]]
 name = "clap_complete"
-version = "4.5.18"
+version = "4.5.26"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


nardoor:
similar / complementary change in rustic-rs/rustic_core#257

aawsome (approved):
LGTM

Issue #251, opened 2022-10-07
`check --read-data`: Add options to check subset of packs + warm-up options


aawsome:
see #250

aawsome:
The following options might be nice to have:

-    only tree pack files (this could be standard as tree pack files should be always readable)
-    only a fixed subset of the packfiles (e.g. starting with given prefixes)
-    a random subset of the packfiles
-    only the packfiles which are referenced by one or a set of snapshots

PR #1230, opened 2024-09-08
feat(check): Allow to only check trees+packs for given snapshots

aawsome:
see #251


simonsan:
Unblocked after merging #1282

src/commands/check.rs
```diff
@@ -27,7 +31,16 @@ impl CheckCmd {
     fn inner_run(&self) -> Result<()> {
         let config = RUSTIC_APP.config();
         let repo = open_repository(&config.repository)?;
-        repo.check(self.opts)?;
+
+        let groups = repo.get_snapshot_group(&self.ids, SnapshotGroupCriterion::new(), |sn| {
```
aawsome, line 35: I think there is also get_matching_snapshots which is better here?!?

simonsan: Pushed some changes, did you mean it like that?

aawsome: Uh, no. I realize that we have `self.ids` - this must be why I used `get_snapshot_group`... I think then it is best to let it as it was... sorry for leading into the wrong direction..

simonsan: No problem ;-)

simonsan (approved):

PR #1241, opened 2024-09-17
fix: webdav/forget: correctly use application config

### LINKED ISSUE #1163: WebDAV configuration isn't working in windows
Tested with `latest` and `rustic v0.7.0-45-gfc8e3c5-nightly`

`rustic webdav` binds to `http://[::1]:8000` despite the configuration.

```
[webdav]
address = "127.0.0.1:8000"
path-template = "[{hostname}]/[{label}]/{time}" # The path template to use for snapshots. {id}, {id_long}, {time}, {username}, {hostname}, {label}, {tags}, {backup_start}, {backup_end} are replaced. [default: "[{hostname}]/[{label}]/{time}"]. Only relevant if no snapshot-path is given.
time-template = "%Y-%m-%d_%H-%M-%S"
symlinks = false
file-access = "read"
```

nardoor:
The `inner_run` implementation used `self`
to read its config when it should use `RUSTIC_APP.config()`.

Some might say this is counterintuitive.
fixes #1163
also now correctly respects having `prune = true` in the config profile.

### DIFF
--- a/src/commands/forget.rs
+++ b/src/commands/forget.rs
@@ -98,6 +98,9 @@ impl Runnable for ForgetCmd {
 }
 
 impl ForgetCmd {
+    /// be careful about self VS RUSTIC_APP.config() usage
+    /// only the RUSTIC_APP.config() involves the TOML and ENV merged configurations
+    /// see https://github.com/rustic-rs/rustic/issues/1242
     fn inner_run(&self) -> Result<()> {
         let config = RUSTIC_APP.config();
         let repo = open_repository(&config.repository)?;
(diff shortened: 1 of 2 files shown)


nardoor:
@aawsome @simonsan 
The logic that caused this bug is very counterintuitive to me.
This bug happened once, and nothing in the type system (compile time) will prevent anyone to re-implement it later on.

I wish `abscissa_core` had some way of **mutating** structures like `WebdavCmd` instead of having the truthful config in the `RUSTIC_APP` static.

aawsome (dismissed):

PR #1247, opened 2024-09-21
fix: ls: Remove printing trailing space

varqox:
Paths are separated by a newline, printing space only hinders parsing. Also paths are now displayed using `Path::display()`.

### DIFF
--- a/src/commands/ls.rs
+++ b/src/commands/ls.rs
@@ -141,7 +141,7 @@ impl LsCmd {
             if self.long {
                 print_node(&node, &path, self.numeric_id);
             } else {
-                println!("{path:?}");
+                println!("{}", path.display());
             }
         }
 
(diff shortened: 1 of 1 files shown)


aawsome:
Hey @varqox Thanks for the PR!

I agree that we should also use `path.display()` here.

Moreover another suggestion: Why not (additionally) add a `--json` option? This would allow for much better parsing, actually... About the json content, you could have a look at what restic currently offes...

aawsome:
And https://github.com/rustic-rs/rustic/issues/1239 might be actually also a good side-catch, in case you are willing to also tackle that ;-)

varqox:
> Hey @varqox Thanks for the PR!
> 
> I agree that we should also use `path.display()` here.

I will patch the PR for that.
> Moreover another suggestion: Why not (additionally) add a `--json` option? This would allow for much better parsing, actually... About the json content, you could have a look at what restic currently offers...

> And #1239 might be actually also a good side-catch, in case you are willing to also tackle that ;-)

Let's make the changes in this PR related to the title, and other in other PRs. I can work on the other issues.

varqox:
Okay, I applied the changes, you can merge now :)

aawsome (approved):
LGTM

PR #1263, opened 2024-09-26
feat: add more filter

aawsome:
Adds the following filter options:
- `filter-parhs-exactly` - filters exactly the given pathlists (supersets don't match)
- `filter-tags-exactly` - filters exactly the given taglists (supersets don't match)
- `filter-before` - filters by time
- `filter-after` - filters by time
- `filter-size` - filters by size of snapshot'ed data (upper and lower limit may be given)
- `filter-size-add` - filters by size added to the repository (upper and lower limit may be given)


config/full.toml
```diff
@@ -67,10 +67,16 @@ throttle = "10kB,10MB" # limit and burst per second; only opendal backends; Defa
...
+filter-hosts = ["host1", "host2"] # Default: []
 filter-labels = ["label1", "label2"] # Default: []
 filter-tags = ["tag1,tag2", "tag3"] # Default: []
+filter-tags-exact = ["tag1,tag2", "tag2"] # Default: []
 filter-paths = ["path1", "path2,path3"] # Default: []
+filter-paths-exact = ["path1", "path2,path3"] # Default: []
+filter-after = "2024-01-01" # Default: not set
+filter-before = "2024-02-05 12:15" # Default: not set
+filter-size = "200MiB" # Default: not set
+filter-size-added = "1 MB..10MB" # Default: not set
```
simonsan, line 79:
```suggestion
filter-size-added = "1MB..10MB" # Default: not set
```

simonsan: Would it work with a whitespace? Would be good to add that to the test cases then

src/filtering.rs
simonsan, line 1: Needs to be adapted to `conflate`

src/filtering.rs
```diff
@@ -116,9 +155,113 @@ impl SnapshotFilter {
...
+        f.write_str("..")?;
+        if let Some(to) = self.to {
+            f.write_str(&to.to_string_as(true))?;
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
```
simonsan, line 247: I think it could be nice to also implement some tests for parsing `BeforeDate` and `AfterDate` to make sure we handle edge cases right.

config/full.toml
aawsome: It should also work with whitspaces..

src/filtering.rs
aawsome: done.

src/filtering.rs
```diff
@@ -118,9 +161,114 @@ impl SnapshotFilter {
             }
         }
 
+        if matches!(&self.filter_after, Some(after) if snapshot.time <= after.0)
+            || matches!(&self.filter_before, Some(before) if snapshot.time >= before.0)
+            || matches!((&self.filter_size,&snapshot.summary), (Some(size),Some(summary)) if !size.matches(summary.total_bytes_processed))
+            || matches!((&self.filter_size_added,&snapshot.summary), (Some(size),Some(summary)) if !size.matches(summary.data_added))
+        {
+            return false;
+        }
```
simonsan, line 170: Can we add a comment here what the implied logic should be, so I can review against it?

src/filtering.rs
```diff
@@ -118,9 +161,114 @@ impl SnapshotFilter {
...
+
         snapshot.paths.matches(&self.filter_paths)
+            && (self.filter_paths_exact.is_empty()
+                || self.filter_paths_exact.contains(&snapshot.paths))
             && snapshot.tags.matches(&self.filter_tags)
+            && (self.filter_tags_exact.is_empty()
+                || self.filter_tags_exact.contains(&snapshot.tags))
             && (self.filter_hosts.is_empty() || self.filter_hosts.contains(&snapshot.hostname))
             && (self.filter_labels.is_empty() || self.filter_labels.contains(&snapshot.label))
```
simonsan, line 179: Same here, a comment with the implied logic would be nice, so we can review against that. Especially because `self.filter_paths_exact.is_empty()` is not negated, which confused me.

src/filtering.rs
```diff
@@ -118,9 +161,114 @@ impl SnapshotFilter {
...
+#[derive(Debug, Clone)]
+struct SizeRange {
+    from: Option<ByteSize>,
+    to: Option<ByteSize>,
+}
+
+impl SizeRange {
+    fn matches(&self, size: u64) -> bool {
+        !matches!(self.from, Some(from) if size < from.0)
+            && !matches!(self.to, Some(to) if size > to.0)
```
simonsan, line 216: Same here about the logics

src/filtering.rs
aawsome: comments are added.

src/filtering.rs
aawsome: added

src/filtering.rs
aawsome: added

simonsan (approved):

PR #1273, opened 2024-09-30
build: reintroduce stack unwinding, don't abort on panic

simonsan:
We have aborting on panic in our rustic `Cargo.toml`. For `rustic` this may affect the stack traces that `human-panic` spits out. This will make it easier, to find bugs, that are reported to us by the users, having their stack traces. Reintroducing the unwinding code will impact the binary size, which is currently around 8 MiB, and it may go up a few MiBs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -148,7 +148,6 @@ lto = "fat"
 debug-assertions = false
 codegen-units = 1
 strip = true
-panic = "abort"
 
 [profile.test]
 opt-level = 1


aawsome:
@simonsan Do you have numbers how much the binary size increases? I mean I would not judge this important enough for a significant size increase, actually..

simonsan:
On Windows:

with `panic = "abort"`: 18.3 MiB
without: 27.7 MB

aawsome:
> On Windows:
> 
> with `panic = "abort"`: 18.3 MiB without: 27.7 MB

puh.. that's heavy! I would actually prefer to leave it as "abort"...

simonsan:
> > On Windows:
> > with `panic = "abort"`: 18.3 MiB without: 27.7 MB
> 
> puh.. that's heavy! I would actually prefer to leave it as "abort"...

What if we activate `abort`, when rustic is `v1.0.0` and we have fixed most of the errors and squashed bugs. I think at this beta-stage a better stack trace is much more valuable and important than the actual binary size. 😬

PR #1276, opened 2024-10-01
feat(commands): add a `docs` command to easily access the user, dev and config documentation

simonsan:


src/commands/docs.rs
```diff
@@ -0,0 +1,66 @@
...
+    #[clap(short, long, group = "documentation")]
+    user: bool,
+
+    /// Open the development documentation
+    #[clap(short, long, group = "documentation")]
+    dev: bool,
+
+    /// Open the config documentation
+    #[clap(short, long, group = "documentation")]
+    config: bool,
```
aawsome, line 31: I think we should not use flags `--user` etc here but just an optional argument such that it can be called as `rustic docs user` or `rustic docs dev`.

aawsome:
Good idea to add such a command!

src/commands/docs.rs
simonsan: Yes, fixed it! 👍🏽

aawsome (approved):
LGTM

PR #1277, opened 2024-10-01
fix(docs/cli): improve the descriptions of the CLI commands

simonsan:
Fixed the capitalization of `dump` for the dump command, and then figured, that there might be some better descriptions for other commands as well.


src/commands.rs
```diff
@@ -88,29 +88,28 @@ enum RusticCmd {
...
+    /// Copy snapshots to other repositories
     Copy(CopyCmd),
 
-    /// Compare two snapshots/paths
-    /// Note that the exclude options only apply for comparison with a local path
+    /// Compare two snapshots or paths
     Diff(DiffCmd),
 
-    /// dump the contents of a file in a snapshot to stdout
+    /// Dump the contents of a file within a snapshot to the terminal
```
aawsome, line 97: Actually it is stdout and it is a very common use-case to use e.g. `rustic dump | mysql`

src/commands.rs
```diff
@@ -88,29 +88,28 @@ enum RusticCmd {
...
     Dump(DumpCmd),
 
-    /// Find in given snapshots
+    /// Find patterns in given snapshots
     Find(FindCmd),
 
-    /// Remove snapshots from the repository
+    /// Mark snapshots to be removed from the repository
```
aawsome, line 103: this removes snapshot(file)s without marking them.

src/commands.rs
```diff
@@ -125,17 +124,17 @@ enum RusticCmd {
...
-    /// Update to the latest rustic release
+    /// Update to the latest stable rustic release
     #[cfg_attr(not(feature = "self-update"), clap(hide = true))]
     SelfUpdate(SelfUpdateCmd),
 
     /// Remove unused data or repack repository pack files
     Prune(PruneCmd),
 
-    /// Restore a snapshot/path
+    /// Restore a snapshot or path
```
aawsome, line 134: this allows to restore a complete snapshot (`restore <SNAPID>`) or a path within a snapshot (`restore <SNAPID>:<PATH>`). Maybe something like `Restore (a path within) a snapshot`?

aawsome (changes requested):

src/commands.rs
simonsan: Yeah, I felt stdout is a relatively technical term, so I thought about making it easier to grasp for newcomers. Wouldn't mind changing it back to `stdout`, it's just less inclusive.

src/commands.rs
simonsan: I thought this was just the case when using `--instant-delete`, otherwise `prune` would delete them?

src/commands.rs
simonsan: I used now both? So it's clear it's on the terminal but also on stdout if someone wants to dig deeper? Is that sufficient?

src/commands.rs
simonsan: Would there be ever snapshots deleted by just running `rustic forget` with never calling `rustic prune`?

aawsome: `forget` removes snapshotfiles directly (unless called with `--dry-run`). And this is basically completely independent from `prune`.
The `--instant-delete` option is only a prune option, for `forget` this is only used if we call `forget --prune` - which is just a shortcut for `rustic forget; rustic prune`.

src/commands.rs
aawsome: when I think about it, we should rather add a warning (here or in the docu)  that this may produce non-printable characters, so it is advisable to redirect stdout...

simonsan: Ok, reverted to `stdout`

src/commands.rs
simonsan: Ah, TIL! Good, reverted it then!

aawsome (approved):
LGTM

PR #1282, opened 2024-10-02
fix(deps): update rustic_core and other dependencies and fix merge precedence

### LINKED ISSUE #1226: RUSTSEC-2024-0370: proc-macro-error is unmaintained
> proc-macro-error is unmaintained

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unmaintained                |
| Package             | `proc-macro-error`                      |
| Version             | `1.0.4`                   |
| URL                 | [https://gitlab.com/CreepySkeleton/proc-macro-error/-/issues/20](https://gitlab.com/CreepySkeleton/proc-macro-error/-/issues/20) |
| Date                | 2024-09-01                         |

proc-macro-error&#39;s maintainer seems to be unreachable, with no commits for 2 years, no releases pushed for 4 years, and no activity on the GitLab repo or response to email.

proc-macro-error also depends on `syn 1.x`, which may be bringing duplicate dependencies into dependant build trees.

## Possible Alternative(s)

- [proc-macro-error2](https://crates.io/crates/proc-macro-error2)

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2024-0370.html) for additional details.

simonsan:
Due to changes in the way how merging of values were applied to configuration values in `BackendOptions` this should is considered a fix. Now values from the configuration file don't overwrite values from environment variables, any more. As it was intended and stated in the documentation.

Please make sure, your workflow is still working as intended, especially that the configuration values and environment variables are applied as you wanted it and as we documented it in the Configuration documentation.

Precedence also applies when inheriting options from other profiles using `use-profiles` or when using multiple profiles, e.g. `rustic -P one -P two`. If the same repository option is defined in `one` and `two` the value of `one` is taken. Before the fix, the value of `two` was taken. But the fix only applies to the repository options. For other options, `one` always had precedence.

Fixes #1226

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -12,10 +12,10 @@ use abscissa_core::config::Config;
 use abscissa_core::path::AbsPathBuf;
 use abscissa_core::FrameworkError;
 use clap::{Parser, ValueHint};
+use conflate::Merge;
 use directories::ProjectDirs;
 use itertools::Itertools;
 use log::Level;
-use merge::Merge;
 use rustic_backend::BackendOptions;
 use rustic_core::RepositoryOptions;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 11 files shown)


simonsan:
Breaking changes documentation here: https://github.com/rustic-rs/docs/pull/92

aawsome:
IMO this is not breaking

aawsome (commented):
LGTM, but the breaking discussion needs to be closed before merging this ;-)

src/config.rs
```diff
@@ -168,16 +170,10 @@ pub struct GlobalOptions {
 
     /// List of environment variables to set (only in config file)
     #[clap(skip)]
-    #[merge(strategy = extend)]
+    #[merge(strategy = conflate::hashmap::overwrite)]
```
simonsan, line 173:
```suggestion
    #[merge(strategy = conflate::hashmap::ignore)]
```

aawsome (approved):
LGTM

PR #1293, opened 2024-10-04
ci: add missing feature self-update to tests and checks

simonsan:
Currently, we are not testing with all features, as `self-update` is missing. This PR adds it back.

### DIFF
--- a/.github/workflows/cross-ci.yml
+++ b/.github/workflows/cross-ci.yml
@@ -28,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         rust: [stable]
-        feature: [default, self-update]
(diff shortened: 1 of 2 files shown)


.github/workflows/cross-ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         rust: [stable]
-        feature: [default]
+        feature: [default, self-update]
```
simonsan, line 31: This is quite annoying, as this will create more runs overall. The alternative is, that I change the `cross-ci` action, to determine, that if it checks `rustic`, to add this feature.

PR #1307, opened 2024-10-09
build: introduce a new feature 'release' that includes the 'self-update' feature

simonsan:
This is intended to be used in our CI/CD pipelines, so we don't need to add 'self-update' feature manually, as that is error-prone.


aawsome (approved):
LGTM

PR #1328, opened 2024-10-19
feat(commands): Add tar output to dump command

aawsome:
The `dump` command will now produce a tar when called with a directory.


aawsome:
is rebased on top of #1329

simonsan (approved):

PR #1339, opened 2024-10-24
feat(commands): More dump options

aawsome:
Adds output as  targz and zip.
Also adds the options `--archive` to choose the ouput format and `--file` to directly specify a file to dump into. When a file is specified, the ouput format is automatically chosen from the file extension, if given.


Cargo.toml
```diff
@@ -106,6 +106,8 @@ open = "5.3.0"
 self_update = { version = "0.39.0", default-features = false, optional = true, features = ["rustls", "archive-tar", "compression-flate2"] } # FIXME: Downgraded to 0.39.0 due to https://github.com/jaemk/self_update/issues/136
 tar = "0.4.42"
 toml = "0.8"
+flate2 = "1.0.34"
+zip = { version = "2.2.0", default-features = false, features = ["deflate", "chrono"] }
```
simonsan, line 110:
```suggestion
flate2 = "1"
zip = { version = "2", default-features = false, features = ["deflate", "chrono"] }
```

Can we use the major version here, I think for anything `>=1.y.z` we should be fine to just use major in our manifest.

PR #1344, opened 2024-10-31
chore: clarify `--use-profile` command in config by using long form

tshepang:


simonsan:
I'm not against it in general, as we should make `rustic` easily approachable. Though running `rustic help` and checking for `-P` which is the first option in `Global` or checking the [config readme](https://github.com/rustic-rs/rustic/tree/main/config#global-options-global) would help. What I think we should probably take as a general rule from it is: `Use long versions of command options in public documentation.`

simonsan (approved):

simonsan:
Forgot to say, thank you for the contribution! 🖖

PR #1348, opened 2024-11-09
chore(build): use Cross.toml

simonsan:


simonsan:
Solved it differently, via the justfile approach. For now no `Cross.toml` is needed.

PR #1353, opened 2024-11-11
refactor(mount): rename fields for clarity

simonsan:


Cargo.toml
```diff
@@ -48,7 +48,7 @@ rustdoc-args = ["--document-private-items", "--generate-link-to-definition"]
 [dependencies]
 abscissa_core = { version = "0.8.1", default-features = false, features = ["application"] }
 rustic_backend = { version = "0.5.0", features = ["cli"] }
-rustic_core = { version = "0.6.0", features = ["cli"] }
+rustic_core = { version = "0.6.1", features = ["cli"] }
```
aawsome, line 51: can we do this in a separate PR?

aawsome: Ah, this is because of `FilePolicy`, right?

simonsan: yes

src/commands/mount.rs
```diff
@@ -31,35 +32,53 @@ pub struct MountCmd {
...
 impl Override<RusticConfig> for MountCmd {
     // Process the given command line options, overriding settings from
     // a configuration file using explicit flags taken from command-line
     // arguments.
     fn override_config(&self, mut config: RusticConfig) -> Result<RusticConfig, FrameworkError> {
-        let mut self_config = self.clone();
-        // merge "mount" section from config file, if given
-        self_config.merge(config.mount);
+        // Merge by precedence, cli <- config <- default
+        let mut self_config = self.clone().merge_precedence(config.mount, Self::new());
```
aawsome, line 66: I just wonder if `self.clone().merge_from(config.mount).merge_from(Self::new())` would not be even more clear?!?

src/commands/mount.rs
```diff
@@ -31,35 +32,53 @@ pub struct MountCmd {
...
+
+        // Other values
+        if self_config.mount_point.is_none() {
+            return Err(ParseError
+                .context("Please specify a valid mount point!")
+                .into());
+        }
+
+        if self_config.options.is_empty() {
```
aawsome, line 75: I think this is not needed as we already have this in the default config

src/commands/mount.rs
```diff
@@ -78,24 +97,32 @@ impl Runnable for MountCmd {
 }
 
 impl MountCmd {
+    fn new() -> Self {
```
aawsome, line 100: I would prefer a different name: `default_config()`. As this is not `pub` currently, we are actually totally free to choose any name ;-)

aawsome (approved):
LGTM

PR #1358, opened 2024-11-18
chore: add max-perf-at-small-bin profile

simonsan:
Results in 38.5% size reduction on Windows 11, and 33% less on openSuse.

We could also use it in `release`?


aawsome:
I think we should have performance measurements before we change this for release..

Is there a big difference between using `"s"` ans `"z"`? If not, I'd prefer using `"s"`..

PR #1360, opened 2024-11-19
refactor: use BTreeMap for env in global options

simonsan:
Signed-off-by: simonsan <14062932+simonsan@users.noreply.github.com>


src/config.rs
```diff
@@ -83,6 +86,15 @@ pub struct RusticConfig {
     pub webdav: Option<Value>,
 }
 
+impl Display for RusticConfig {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let config =
+            toml::to_string_pretty(self).unwrap_or_else(|_| "Error serializing config".to_string());
```
nardoor, line 92:
```suggestion
            toml::to_string_pretty(self).unwrap_or_else(|_| "<Error serializing config>".to_string());
```

This might be an error case we **never** encounter.
Anyway I propose some way to clarify that the error case in the display is some <placeholder>.

nardoor (approved):
Lgtm !

I just proposed a small change, most likely unnecessary in the `impl Display`

PR #1364, opened 2024-11-23
feat: shut down gracefully with ctrl+c

simonsan:
If you have a long-running process, like `webdav` or `backup` or others, pressing CTRL+C usually results in a non-zero exit code:

`process didn't exit successfully: P:\CARGO\.cargo-target-win\debug\rustic.exe -P <PROFILE> webdav (exit code: 0xc000013a, STATUS_CONTROL_C_EXIT)`

This shouldn't be the case, as it was user initiated and we can shut down gracefully.

This PR adds this functionality, so CTRL+C shuts down `rustic` gracefully and exits with a 0 exit-code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1114,6 +1114,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "ctrlc"
+version = "3.4.5"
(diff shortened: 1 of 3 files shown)


aawsome (approved):
LGTM

PR #1372, opened 2024-11-30
feat: Add --filter-jq option

aawsome:
Adds filtering using `jq` syntax. You can do something like `--filter-jq '.summary.files_new > 51'` - works similar to what you'd expect from `jq`.

Also puts the `rhai` and `jq` filtering into features which are enabled by default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -121,10 +121,10 @@ tar = "0.4.43"
 toml = "0.8"
 
 # filtering
-rhai = { version = "1", features = ["sync", "serde", "no_optimize", "no_module", "no_custom_syntax", "only_i64"], optional = true }
 jaq-core = { version = "2.0.0", optional = true }
-jaq-std = { version = "2.0.0", optional = true }
 jaq-json = { version = "1.0.0", features = ["serde_json"], optional = true }
+jaq-std = { version = "2.0.0", optional = true }
+rhai = { version = "1", features = ["sync", "serde", "no_optimize", "no_module", "no_custom_syntax", "only_i64"], optional = true }
 
 [dev-dependencies]
 abscissa_core = { version = "0.8.1", default-features = false, features = ["testing"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -121,9 +121,9 @@ tar = "0.4.43"
 toml = "0.8"
 
 # filtering
-jaq-core = { version = "2.0.0", optional = true }
-jaq-json = { version = "1.0.0", features = ["serde_json"], optional = true }
-jaq-std = { version = "2.0.0", optional = true }
+jaq-core = { version = "2", optional = true }
+jaq-json = { version = "1", features = ["serde_json"], optional = true }
+jaq-std = { version = "2", optional = true }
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -117,6 +120,12 @@ self_update = { version = "=0.39.0", default-features = false, optional = true,
 tar = "0.4.43"
 toml = "0.8"
 
+# filtering
+jaq-core = { version = "2.0.0", optional = true }
+jaq-json = { version = "1.0.0", features = ["serde_json"], optional = true }
+jaq-std = { version = "2.0.0", optional = true }
```
simonsan, line 126:
```suggestion
jaq-core = { version = "2", optional = true }
jaq-json = { version = "1", features = ["serde_json"], optional = true }
jaq-std = { version = "2", optional = true }
```

src/filtering.rs
```diff
@@ -62,6 +70,75 @@ impl SnapshotFn {
...
+        let programm = File { code: s, path: () };
+        let loader = Loader::new(jaq_std::defs().chain(jaq_json::defs()));
+        let arena = Arena::default();
+        let modules = loader
+            .load(&arena, programm)
+            .map_err(|errs| anyhow!("errors in jq: {errs:?}"))?;
+        let filter = Compiler::<_, Native<_>>::default()
+            .with_funs(jaq_std::funs().chain(jaq_json::funs()))
+            .compile(modules)
+            .map_err(|errs| anyhow!("errors in jq: {errs:?}"))?;
```
simonsan, line 102:
```suggestion
            .map_err(|errs| anyhow!("errors during compiling filters in jq: {errs:?}"))?;
```

src/filtering.rs
```diff
@@ -62,6 +70,75 @@ impl SnapshotFn {
...
+#[cfg(feature = "jq")]
+impl FromStr for SnapshotJq {
+    type Err = anyhow::Error;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let programm = File { code: s, path: () };
+        let loader = Loader::new(jaq_std::defs().chain(jaq_json::defs()));
+        let arena = Arena::default();
+        let modules = loader
+            .load(&arena, programm)
+            .map_err(|errs| anyhow!("errors in jq: {errs:?}"))?;
```
simonsan, line 98:
```suggestion
            .map_err(|errs| anyhow!("errors loading modules in jq: {errs:?}"))?;
```

src/filtering.rs
```diff
@@ -143,6 +228,7 @@ impl SnapshotFilter {
     /// `true` if the snapshot matches the filter, `false` otherwise
     #[must_use]
     pub fn matches(&self, snapshot: &SnapshotFile) -> bool {
+        #[cfg(feature = "rhai")]
```
simonsan, line 231: See discord, maybe it makes sense to make jq and rhai filter options mutually exclusive?
Or what is the intended way that both interact with each other when both are set?

simonsan (dismissed):

simonsan (approved):

PR #1378, opened 2024-12-04
fix(config): set a non-zero default progress interval for progress options

aawsome:


src/config/progress_options.rs
```diff
@@ -42,7 +48,8 @@ impl ProgressOptions {
     ///
     /// `Duration::ZERO` if no progress is enabled
```
simonsan, line 49:
```suggestion
    /// `Duration::ZERO` if no progress is enabled
```

Does that need an update as well?

aawsome: is changed.

simonsan (approved):

PR #1385, opened 2024-12-10
feat(commands): Add missing key subcommands

### LINKED ISSUE #983: `key`: better key handling and additional sub commands
```
$ rustic --version
rustic v0.6.1
```

```
$ rustic key list
error: unrecognized subcommand 'list'

Usage: rustic key [OPTIONS] <COMMAND>

For more information, try '--help'.
```

```
$ rustic key --help
Manage keys

Usage: rustic key [OPTIONS] <COMMAND>

Commands:
  add   Add a new key to the repository
  help  Print this message or the help of the given subcommand(s)

Options:
  -h, --help
          Print help (see a summary with '-h')
```

Looks like `key list` command has got lost during latest refactoring...

Unless documentation is wrong:
https://rustic.cli.rs/docs/commands/misc/key.html

Possibly also `remove` and `passwd`

aawsome:
closes #983

### DIFF
--- a/src/commands/key.rs
+++ b/src/commands/key.rs
@@ -204,7 +204,11 @@ impl Runnable for PasswordCmd {
 impl PasswordCmd {
     fn inner_run(&self, repo: CliOpenRepo) -> Result<()> {
         let pass = self.pass_opts.pass("enter new password")?;
-        let key_opts = KeyOptions::default();
+        let old_key: KeyFile = repo.get_file(repo.key_id())?;
+        let key_opts = KeyOptions::default()
+            .hostname(old_key.hostname)
+            .username(old_key.username)
+            .with_created(old_key.created.is_some());
         let id = repo.add_key(&pass, &key_opts)?;
         info!("key {id} successfully added.");
 
--- a/src/commands/key.rs
+++ b/src/commands/key.rs
@@ -1,15 +1,17 @@
 //! `key` subcommand
 
-use crate::{Application, RUSTIC_APP, repository::CliOpenRepo, status_err};
+use crate::{
+    Application, RUSTIC_APP, helpers::table_with_titles, repository::CliOpenRepo, status_err,
+};
 
 use std::path::PathBuf;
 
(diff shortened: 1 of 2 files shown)


src/commands/key.rs
```diff
@@ -63,29 +127,97 @@ impl Runnable for AddCmd {
...
+        };
+    }
+}
+
+impl RemoveCmd {
+    fn inner_run(&self, repo: CliOpenRepo) -> Result<()> {
+        repo.delete_key(&self.id)?;
+        info!("key {} successfully removed.", self.id);
+        Ok(())
+    }
```
simonsan, line 182: There is no safe-guarding in place, e.g. a second question as in 'Key {key_id} exists, do you really want to irrecoverably delete this key? (y/N)' with defaulting to `No`.

simonsan: That's good, and some safe-guarding against locking oneself out. But I think it's still important to add a normal feedback loop for deleting something. We could add a `--force` delete, so that feedback loop is being jumped over.

aawsome: with `forget` and `prune` we also don't have a feedback loop, so I would find it very irritating to have one here. `forget` and `prune` do have `--dry-run`, but as you specify a key id to remove, I don't see much value in adding a dry-run option here..

simonsan: To be fair, I think it's a bit different with `forget` and `prune`, as you also said, and they even have a `dry-run` option. I think it's just good design, to have a feedback loop when deleting something, and optionally make that circumventable. I would find it pretty weird to not have that, to be honest. The loop itself is also important to give feedback on if the key was found at all and what happens to it. I think the whole `key` subcommand(s) make a pretty good role model for a `tui`, actually. Where there would be probably also a feedback loop when deleting.

simonsan: https://docs.rs/dialoguer/0.11.0/dialoguer/struct.Confirm.html

aawsome:
rebased to current main

Issue #1391, opened 2024-12-30
Cannot set snake case hot/cold repo options via environment variables


k9withabone:
I'm trying to set the `account_key` hot/cold repository option for the OpenDAL azblob backend via an environment variable so that I can load it from a systemd credential. This seems to be impossible at the moment as `RUSTIC_REPO_OPTHOT_*` and `RUSTIC_REPO_OPTCOLD_*` variables are converted to kebab case. #1081 allowed for setting `OPENDAL_*` variables as snake case, but those options then apply to both the hot and cold repository.

Rustic version: 0.9.5

aawsome:
Thanks for opening this issue @k9withabone 
I created #1392 which should solve this.

PR #1393, opened 2025-01-02
fix: rename OPENDAL_HOT_*/COLD_* to OPENDALHOT_*/COLD_*

aawsome:
#1392 was unfortunately ambiguous. IMO the better option to solve #1391 is to use OPENDALHOT_* and OPENDALCOLD_* env variables.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -216,7 +216,7 @@ impl Configurable<RusticConfig> for EntryPoint {
         // That's why it says `_config`, because it's not read at all and therefore not needed.
         let mut config = self.config.clone();
 
-        // collect "RUSTIC_REPO_OPT*" and "OPENDAL_*" env variables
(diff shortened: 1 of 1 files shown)


aawsome:
It is actually a bit complicated. The opendal project also uses `OPENDAL_*` variables in their own CLI (I think they call it `oli`), so it might be beneficial to also use those. Besides this, you are right that most env variables we use start with `RUSTIC_*` - however, these are all the ones where we have a CLI option which is replaced by this. Here we define options which are not exposed as CLI parametes, but only within the config profile..

Anyway, the amiguity if `OPENDA_HOT_XXX=YYY` should be interpreted as the parameter `xxx=yyy` for a hot repo repo (which we want here)  or as `hot_xxx=yyy` for hot/cold part (which is what is used without this PR) must be solved, IMO.

PR #1404, opened 2025-01-20
feat: push prometheus metrics to pushgateway on-demand

### LINKED ISSUE #1403: JSON output as prometheus metric format?
Hey!

I've been using rustic for a bit, and I recently submitted [a module for easily using it to NixOS](https://github.com/NixOS/ nixpkgs/pull/373042) (my personal distribution of choice).

As part of it, I added one thing useful at least to me: parsing the JSON output of `--json` and turning it into prometheus metrics that could then be exported via the node-exporter.

But upon thinking about it some more I'm curious: are you interested in upstreaming this work, so that `rustic` could gain a `--prometheus` flag that'd output the result as prometheus metrics, rather than raw JSON?

TBH the JSON currently output is not very nice to work with, eg. because parsing times (esp. when TZ is not UTC) is hard — hence the hacks used in the NixOS module's script to work around it.

WDYT?

Anyway, thank you for all you're doing with `rustic`!

Ekleog:
Fixes #1403


src/commands/backup.rs
```diff
@@ -72,6 +72,27 @@ pub struct BackupCmd {
...
+    #[cfg(feature = "prometheus")]
+    #[clap(long, value_name = "PUSHGATEWAY_URL", value_hint = ValueHint::Url)]
+    #[merge(strategy=conflate::option::overwrite_none)]
+    prometheus: Option<String>,
+
+    /// Job name for the Pushgateway push
+    #[cfg(feature = "prometheus")]
+    #[clap(long, value_name = "JOB_NAME", default_value = "rustic")]
+    #[merge(skip)]
+    prometheus_job: String,
```
aawsome, line 88: with `merge(skip)`, this won't be configurable via the config profiles. Better also use a `Option<String>` with `overwrite(none)` merge strategy and define a default value for `None`.

src/prometheus.rs
```diff
@@ -0,0 +1,189 @@
+use std::{env, error::Error};
+
+use prometheus::register_gauge;
+use rustic_core::repofile::SnapshotFile;
+
+pub fn parse_label(
+    s: &str,
+) -> Result<(String, String), Box<dyn Error + Send + Sync + 'static>> {
```
aawsome, line 8: I wonder if we can use clap for parsing this, but if not, please use `anyhow::Result` here..

src/prometheus.rs
```diff
@@ -0,0 +1,189 @@
...
+) -> anyhow::Result<()> {
+    let summary = snap.summary.as_ref().expect("Reaching the 'push to prometheus' point should only happen for successful backups, which must have a summary set.");
+
+    let metric_time = register_gauge!("rustic_backup_time", "Timestamp of this snapshot",).unwrap();
+    metric_time.set(snap.time.timestamp_millis() as f64 / 1000.);
+    let metric_files_new = register_gauge!(
+        "rustic_backup_files_new",
+        "New files compared to the last (i.e. parent) snapshot",
+    )
+    .unwrap();
```
aawsome, line 29: I think we can just bubble up all errors from the `register_gauge!()` by using `?`.

aawsome (changes requested):
Sorry for the late review.. From my side this is almost ready for merging, I just left a view comments.

General question: What do you think about which env variables are used? The "standard" is to use `RUSTIC_XXX` env variable for a `xxx` option. We have the `pushgateway_url` option and we have (only) env options for `PUSHGATEWAY_USER` and `PUSHGATEWAY_PASS` - I personally would have the later 2 also in the config and then we should decide about the env names. Or use `RUSTIC_PUSHGATEWAY_*` (by clap) if present and the `PUSHGATEWAY_*` as alternatives if they are commonly used variables for the pushgateway?

src/commands/backup.rs
Ekleog: Done :)

src/prometheus.rs
Ekleog: The reason I did it this way is based on [this page of the clap documentation](https://docs.rs/clap/latest/clap/_cookbook/typed_derive/index.html), so I'd assume clap offers no built-in way to do this. I just adjusted to `anyhow::Result`! :)

src/prometheus.rs
Ekleog: Good point! I had done it this way because for some reason prometheus-using people always do that. I also added `.context()` just in case :)

Ekleog:
Thank you for your review!

About the authentication environment variables, I actually thought about it. My argument against having CLI flags with user/pass is, it's essentially insecure to use this way, because any process on your machine can then read the CLI flags, and thus bypass auth.

Especially when Pushgateway is likely often deployed on the same host it's keeping metrics for (and firewalled out of the internet), this seems to me like having its credentials in the CLI would basically completely kill any security the credentials would bring.

I just renamed them `RUSTIC_*` for consistency with clap.

Also, I would not at all be opposed to having these as a parameter in the config file, but TBH I have absolutely no idea how to actually do it without also exposing a clap argument — I haven't dived deep enough in rustic's config handling yet.

WDYT?

aawsome:
Hi @Ekleog - sorry for the late reply...

I overworked this PR a bit:
- now uses the crate `prometheus_push` which allows to fine-tune the used `reqwest`. Especially it allows to build without openssl which may be a block for building on some targets.
- I moved most of the options to `GlobalOptions` as I think prometheus metrics may be also interesting for other commands in future (like `forget`, `prune`, maybe `check`). This also includes moving the push code as a method of this struct.
- Regarding the user/password discussion: We already have the `--password` option which is also not advisable to use. I think we should be consistent here to also have the pushgateway user/pass as CLI arguments.

Can you please test if this still works with my changes?

If you are fine with what I was doing, I think the only open points are to adapt unit tests and docu / config examples accordingly.

Ekleog:
Hmm… for `prometheus-push`, are you really sure you need it for one of your supported targets? It seems to be very young, especially compared to the main `prometheus` crate; and a target with zero native-tls-supported target sounds specific enough to likely not be supported by rustic anyway… I think?

As for your other changes, they definitely make sense to me! I'll review and re-test as soon as you confirm where you want to go wrt. prometheus-push. I personally would not be willing to accept, in a project I maintain, a dependency with a first release 12 months ago, a last release 8 months ago, and that started for some reason at version 0.2.1, all that with less than 10k downloads. But that's your choice, so just let me know your decision and I'll follow! :)

Ekleog:
Also, no worries at all about the reply delay, and thank you for your very involved answer!

aawsome:
Thanks for the discussion  @Ekleog 

I changed it now by copy&pasting few lines from the prometheus codebase which allows to use a self-defined reqwest without any additional dependency.

> Hmm… for `prometheus-push`, are you really sure you need it for one of your supported targets? It seems to be very young, especially compared to the main `prometheus` crate; and a target with zero native-tls-supported target sounds specific enough to likely not be supported by rustic anyway… I think?

Ok, in fact it is more about the need to have something like libssl-dev on the system compiling rustic - and the runtime library then needed to run it. This is all source of potential trouble. And when it come to cross-compiling, things may even go wild..

aawsome:
@Ekleog 

One thing regarding the job name: In case we want to monitor, lets say, `backup` and `prune` both with their own metric, would a use want to have identical job names here (like `rustic-backup`/`rustic-prune`) or would a common name be used (`rustic`)? 
Currently the job name is a global option, if it may differ, it should be rather moved to the command...

And about labels: This is currently done such that users can give labels on the global level and addtional labels on the backup (and even specific to snapshots defined in the config profile) level. I hope this is general enough to suit everybody's need...

Ekleog:
The changes make sense to me!

For the job name, I think (though I haven't checked) that the push gateway will delete old metrics upon getting new metrics with the same job, even if the old metrics was not the same.

So if people want to run backup then prune with no interval, but keep seeing the metrics until the next run of each, we'll need to have a different job name per command.

This being said it could still make sense to have it be in globaloptions. Thinking more, I'm actually thinking that we should not have it have any default value: forcing the user to configure it would solve all the questions, and it'd avoid users with multiple backup runs erasing their own metrics by default.

Does that make sense?

Ekleog:
(as for the global/extra labels, TBH I'm thinking labels should be defined as CLI arguments usually because people may have multiple backup/forget jobs that'd want different labels, so I'd say I'm opinionless on this one)

aawsome:
@Ekleog about the CLI-options:
Of course everything can be given as CLI options, however the question is about how to enable it in the config profile such that users can only run `rustic backup` or `rustic prune` with everything already defined. So, I moved the job name to the backup options and did some finalizing including the unit tests and docu, see the `full.toml` or the Readme for an example how a config profile may look like.

I just encountered one (hopefully) thing: You set the label `path` to the path, but in labels no `/` is allowed, so this can produce errors in most cases. Any idea how we should handle/escape the `/`?

Also, can you please test if everything is working as expected? I didn't setup a prometheus or pushgateway, so I can't really test it.. ;-)

Ekleog:
I'm not sure I get what you mean by "no / allowed"? Prometheus label values [can be any Unicode character](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels)

I'll test it out hopefully by the end of weekend, I have a lot of things to do right now unfortunately! I'll make sure to report here as soon as I have tested it 😄

Ekleog:
Also, thank you very much for all your involvement in this PR! Sorry I've been underwater to answer you over the last few days, I'll do my best to give you a proper test asap ❤️

aawsome:
@Ekleog I changed the labels to use base64 encoding as described in https://github.com/prometheus/pushgateway/blob/master/README.md.

One thing: There are quite some default labels set like paths, uid, gid - and those can't be unset by users. I would actually prefer using a minimal set of labels - especially as users can always define their own labels. Maybe even use no label at all or restrict them to the ones which can be used for grouping: `hostname`, `label` (the rustic label) and `paths`. WDYT?

Also would be nice to have a tests against a running prometheus pushclient before we merge this.

Ekleog:
That's a great idea, to avoid issues related to eg. quotes in paths or other labels' values!

As for limiting the default labels, I can confirm that hostname+label+paths would be enough for my personal use case! I actually only need an `id` label, that I would define myself on the backup, so :shrug: 

Just for the test against a running prometheus pushclient, do you mean a cargo unit test (which seems hard considering it'd require a running pushclient), or a CI job that'd run rustic once and confirm there are expected metrics?

aawsome:
> As for limiting the default labels, I can confirm that hostname+label+paths would be enough for my personal use case! I actually only need an `id` label, that I would define myself on the backup, so 🤷

I restricted the default labels to hostname+label+paths+tags.
 
> Just for the test against a running prometheus pushclient, do you mean a cargo unit test (which seems hard considering it'd require a running pushclient), or a CI job that'd run rustic once and confirm there are expected metrics?

I added a unit test. I think the only thing missing is just one manual test to verify that the format is really what pushclient can work with...

aawsome:
I installed a pushgateway and tested myself - needed a few corrections and added debug and a better testcase.

@Ekleog Only one thing I stumbled over: We are using `Gauge` for all values. I don't have experience with prometheus, but it seems there are other metric types. I wonder if we really want to use `Gauge` for all values or if we might differentiate. E.g. the dates are now saved as seconds since 1970, if I am correct - doesn't prometheus support a datetime metric?

Ekleog (commented):
Oh wow, I'm impressed by how motivated you are on this PR, thank you for all the work you put in here!

For the metrics, [here are the supported Prometheus metric types](https://prometheus.io/docs/concepts/metric_types/). In this case, I think gauge would be the right choice for all the metrics — because it's from a batch job, there's no meaningful "reset to 0" semantics for counters :)

Also, I did try switching my production system to the current tip of this PR (a65f2e4012e6a353014fa5824dfffcb616412862).

Everything seems to be working fine for me! :tada: Though I still need to adjust the exact labels now that everything's going through the pushgateway, which broke my dashboards and alerts — hopefully I'll manage to get something working again soon, but it should not be an issue considering all metrics seem to properly have entered prometheus :)

I'll be very happy to switch back to a released version of rustic as soon as it lands :smile:

Ekleog:
(I added some comments in a review to not touch the code you had written, but I'd be happy to fix them myself if you confirm they sound reasonable to you, next time I find some time to work on my personal backup system! :))

Ekleog:
Ok, I can confirm that a few days after, all my metrics dashboards are adjusted and look the same as before; and backups seem to be working without issues either!

Just let me know if you want me to handle the comments I posted, I don't want to disrupt your workflow by pushing to the branch at an unexpected time :smile:

src/config.rs
```diff
@@ -190,6 +193,134 @@ pub struct GlobalOptions {
...
+            return Ok(());
+        };
+
+        // Merge labels and extra labels
+        let labels = self
+            .prometheus_labels
+            .iter()
+            .map(|x| (x.0.as_str(), x.1.as_str()))
+            .chain(extra_labels.iter().map(|(a, b)| (a.as_str(), b.as_str())))
+            .collect();
```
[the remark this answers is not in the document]
aawsome: Is changed, but I made the hardcoded labels the ones with lowest priority..

src/commands/backup.rs
```diff
@@ -329,7 +341,170 @@ impl BackupCmd {
             println!("snapshot {} successfully saved.", snap.id);
         }
 
+        #[cfg(feature = "prometheus")]
+        if config.global.is_prometheus_configured() {
+            if let Err(err) = publish_metrics(&snap, self.prometheus_job, self.prometheus_labels) {
+                warn!("error pushing prometheus metrics: {err}");
+            }
+        }
```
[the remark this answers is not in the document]
aawsome: I added a warning for the case of configured but not compiled prometheus logging...
Note that we intentionally allow to configure the options even for versions without compiled prometheus support - it would be even more inconvenient to get parsing errors for the config toml...

Ekleog (commented):
That's awesome, thank you!

The PR as-is looks perfect to me, are you missing anything on your end about it? :smile:

aawsome (approved):
LGTM

aawsome:
Thanks a lot @Ekleog for the PR and the reviews of the overworking!

Ekleog:
I'm the one who should say thank you for all the work you put in there! I'll follow #1386 to know when it's released.

Thank you very much for all your work! :purple_heart:

PR #1523, opened 2025-08-08
fix(commands): Skip warmup script on --dry-run

### LINKED ISSUE #1428: `prune --dry-run` issues warm-up command
When using hot/cold storage `rustic prune --dry-run` will issue warmup commands.

```
$ rustic -V
rustic 0.9.5
$ rustic prune --dry-run
[INFO] using config /etc/rustic/rustic.toml
[INFO] using warm-up command /etc/rustic/warmup.sh '%id'
[INFO] repository rclone:rustic-cold:#rclone:rustic-hot:: password is correct.
[INFO] using cache at /root/.cache/rustic/63f34d4a000baa0a279bf2e06728c44ff8a852e184ef0e04d992067bf1f30fd5
[00:00:00] reading index...               ████████████████████████████████████████          7/7                                                  [00:00:00] reading snapshots...           ████████████████████████████████████████          2/2                                                  [00:00:04] finding used blobs...          ████████████████████████████████████████          2/2                                                  [00:00:00] getting packs from repository...                                                                                                      to repack:          3 packs,       2411 blobs,    2.0 MiB
this removes:                        17 blobs,   74.9 kiB
to delete:          0 packs,          0 blobs,        0 B
unindexed:          0 packs,         ?? blobs,        0 B
total prune:                         17 blobs,   74.9 kiB
remaining:                       133080 blobs,    2.7 GiB
unused size after prune:   19.6 MiB (0.71% of remaining size)

packs marked for deletion:          0,        0 B
 - complete deletion:               0,        0 B
 - keep marked:                     0,        0 B
 - recover:                         0,        0 B
[00:00:00] warming up packs...            ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░          0/3                                                  Warmup request for e139768de46015154fe83af4ce4f3c7e562588192b7983c88e3cca37c596f79d
Warmup request for f5f17aee0cf85d8c3833257fcdc42c3d5f894c6c2b4a213cbe7ac30b741078c0
Warmup request for 3c84f56aa7aee92f2791afe8dc6b09b7e1720061b23ac62bacaa9785a5574882
[00:00:00] warming up packs...            ████████████████████████████████████████          3/3            
```

The `Warmup request` lines come from my warmup script.

Configuration file:
```
[repository]
repository = "rclone:rustic-cold:"
repo-hot = "rclone:rustic-hot:"
warm-up-command = "/etc/rustic/warmup.sh %id"
warm-up-wait = "48h"

[forget]
keep-within = "30d"
keep-daily = 180
keep-monthly = 18

[backup]
exclude-if-present = [".nobackup", "CACHEDIR.TAG"]
group-by = "host,label"
one-file-system = true
parent = "latest"
```

SomeThink1729:
Skip running the warmup script when using `--dry-run` for `prune` and `restore`.
The new option `--dry-run-warmup` has been added and can be used to get the prior behavior (i.e. warm-up for a dry-run).

closes #1428

### DIFF
--- a/src/commands/prune.rs
+++ b/src/commands/prune.rs
@@ -44,7 +44,7 @@ impl PruneCmd {
         if dry_run && config.global.dry_run_warmup {
             repo.warm_up(prune_plan.repack_packs().into_iter())?;
         } else if !config.global.dry_run_warmup {
-            info!("Ignoring --dry-run-warmup works only in combination with --dry-run");
+            debug!("Ignoring --dry-run-warmup works only in combination with --dry-run");
         }
         if !dry_run {
             repo.prune(&self.opts, prune_plan)?;
--- a/src/commands/restore.rs
+++ b/src/commands/restore.rs
@@ -6,7 +6,7 @@ use crate::{
 
 use abscissa_core::{Command, Runnable, Shutdown};
 use anyhow::Result;
-use log::info;
+use log::{debug, info};
 
 use rustic_core::{LocalDestination, LsOptions, RestoreOptions};
 
@@ -104,7 +104,7 @@ impl RestoreCmd {
             repo.restore(restore_infos, &self.opts, ls, &dest)?;
             println!("restore done.");
         } else {
-            info!(
+            debug!(
                 "--dry-run is without warmup, --dry-run --dry-run-warmup also issues the warmup script."
             );
         }
--- a/tests/snapshots/show_config__show_config_passes.snap
+++ b/tests/snapshots/show_config__show_config_passes.snap
@@ -21,8 +21,6 @@ run-finally = []
 [global.metrics-labels]
 
 [repository]
-repository = "/tmp/repo"
-password = "test"
 no-cache = false
 warm-up = false
 


aawsome:
@SomeThink1729 Thanks a lot for opening this PR!

I think we should first talk a bit about the general direction. While I see that users may not want to warm-up during a `--dry-run`, there are also others which exactly do want to do this - run a `dry-run` to warm-up some data which can then be processed in future.

So, I suggest that we make warming up in dry run an option to choose, i.e. either add a `--dry-run-warm-up` option along with `--dry-run` or add a `--dry-run-no-warm-up` (names can be discussed) option to keep the current behavior but still allow users to choose.

@SomeThink1729 Which variant do you like more? Can you add this to this PR?

SomeThink1729:
@aawsome Thanks for the feedback.

Until now I didn't see the use case you're describing, but now it totally makes sense. 
I only saw the warm-up script as something to establish a connection - login somewhere else etc. transform data, which you wan't to skip if you `--dry-run`.

For me personally I think this use case is more naturally and I'd put in under `--dry-run` and the one you're describing under `--dry-run-warmup`, but this would lead to a breaking change in `--dry-run`, which I don't really like.

So I'll implement it the other way round (`--dry-run` and `--dry-run-no-warmup`) and we can discuss things further, what you like more etc.
I think changing the names is not that much of an effort after adding the feature.

aawsome:
After thinking about it, I agree that it may make more sense to have dry run not warm up by default. The next release will anyway break smaller things, so I think we can also introduce a breaking change here...

SomeThink1729:
@aawsome I've now added the discussed behaviour. 

`--dry-run` and `--dry-run-warmup` clash with each other, since one should issue the script and the other doesn't.
Right now it takes the stricter one of both, so it skips warmup, but I prefer to do nothing and return an error that this config is invalid. I thought about adding something like this:
```
pub fn validate(&self) -> Result<(), SomeKindOfError> {
// do some checks
}
```
to check if both arguments are supplied or if the config is valid and continue execution.
However I have no clue what kind of error type I would need or where to call this function.

Do you think it's necessary and would like to include such a check or do you think its unnecessary?
Tests are still missing, but I intend to write some before the merge

aawsome:
> `--dry-run` and `--dry-run-warmup` clash with each other, since one should issue the script and the other doesn't. Right now it takes the stricter one of both, so it skips warmup, but I prefer to do nothing and return an error that this config is invalid.

We could let `clap` handle this. However, this is also present in the config file, so it won't work when given there...
As we use `anyhow` for the rustic CLI, that would mean to just return an `anyhow!` error in this case or just use `bail`. IIRC in the `backup` command there are some checks which are handled in this way.

An alternative would be to have a `--dry-run` option and the `--dry-run-warmup` be an option that only yields in combination with `--dry-run`. So, `--dry-run` would not warmup and `--dry-run --dry-run-warmup` would warmup. But using only `--dry-run-warmup` would do nothing or only give a warning that it is ignored as no `dry-run` is given... This however would mean to change the logic a bit...

aawsome:
@SomeThink1729 would you like to continue working on this? If not, I can also finish this to merge it soon...

aawsome (approved):
LGTM! Thanks @SomeThink1729

PR #1529, opened 2025-08-18
docs(config) Correct skip-identical-parent key

grifferz:
`skip-identical-parent` was listed as `skip-if-unchanged`, its corresponding restic command line option.

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -190,41 +190,41 @@ See [Global Hooks](#global-hooks-globalhooks).
 **Note**: If set here, the backup options apply for all sources, although they
 can be overwritten in the source-specific configuration, see below.
 
-| Attribute          | Description                                                                                     | Default Value         | Example Value | CLI Option              |
-| ------------------ | ----------------------------------------------------------------------------------------------- | --------------------- | ------------- | ----------------------- |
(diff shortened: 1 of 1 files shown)


aawsome:
@grifferz Thanks for proposing this PR!

The name of the option was changed in https://github.com/rustic-rs/rustic_core/pull/387, so `skip-identical-parent` is correct. This is not yet released, so only present in the nightly builds.
The change however is not yet done in the "CLI Option" column, so maybe you can correct this?

grifferz:
Sorry, I am confused. main currently lists the config file option as `skip-if-unchanged` whereas it should be `skip-identical-parent`. So that's what I changed it to in this PR. main does currently (correctly) say `--skip-identical-parent` as the CLI option.

Are you saying that in fact the toml file _should_ use `skip-if-unchanged`?

aawsome:
Sorry for the confusion @grifferz.
In fact in main the config option is `skip-if-unchanged` and the CLI option is `--skip-if-unchanged`. In the doc, the first is already correctly changed, but the second was obviously missed and is wrong in the docu.

grifferz:
Right okay so the option actually is `skip-if-unchanged` in both cases and in actual use for me `skip-unchanged-parent` is only working because I am using the most recent release. not a nightly. Got it. Sorry for not investigating this with a nightly.

PR #1531, opened 2025-08-20
feat: Add --name option to select snapshot(s) from config to backup

### LINKED ISSUE #1520: Please add backup "name" or "id" to config
Right now backups are identified by "sources":

```
[[backup.snapshots]]
sources = [ "/home" ]
tags = [ "home" ]
```

So I can run `rustic backup /home` and it'll automatically apply all settings like "home" tag, hooks, etc.

The problem is when I want to say `rustic backup /home` but In fact I need to backup different folder due to btrfs snapshots:

So the idea is to decouple "snapshot" selection from actual source path to backup and get something like:

```
[[backup.snapshots]]
name = "/home"
sources = [ "/mnt/snapshot/home" ]
tags = [ "home" ]
as-path = "/home"
hooks.run_before = [ "btrfs subvolume snapshot -r /home /mnt/snapshot/home" ]
hooks.run_finally = [ "btrfs subvolume delete -c /mnt/snapshot/home" ]
```
Basically I still can do it like this and just run as `rustic backup /mnt/snapshot/home`. But it less convenient. 

Maybe we can name it as `id = @home` and use `rustic backup @home` syntax

aawsome:
closes #1520

### DIFF
--- a/src/commands/backup.rs
+++ b/src/commands/backup.rs
@@ -18,7 +18,7 @@ use abscissa_core::{Command, Runnable, Shutdown};
 use anyhow::{Context, Result, anyhow, bail};
 use clap::ValueHint;
 use comfy_table::Cell;
-use conflate::Merge;
+use conflate::{Merge, MergeFrom};
 use log::{debug, error, info, warn};
 use rustic_core::StringList;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


aawsome:
Actually, I think that using `id` is not optimal here as there are already snapshot ids which are hashes of the `SnapshotFile`. Maybe `name` would be a better fit?

PR #1545, opened 2025-08-29
feat(commands): make group_by a global option

aawsome:
`group-by` can now be given as a global option.
This also allows to set it in the config profiles or use the env variable `RUSTIC_GROUP_BY`.

### DIFF
--- a/src/commands/find.rs
+++ b/src/commands/find.rs
@@ -11,7 +11,7 @@ use globset::{Glob, GlobBuilder, GlobSetBuilder};
 use itertools::Itertools;
 
 use rustic_core::{
-    FindMatches, FindNode, SnapshotGroupCriterion,
+    FindMatches, FindNode,
     repofile::{Node, SnapshotFile},
 };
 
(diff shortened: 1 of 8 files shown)


src/config.rs
```diff
@@ -158,6 +159,18 @@ pub struct GlobalOptions {
     #[merge(strategy=conflate::vec::append)]
     pub use_profiles: Vec<String>,
 
+    /// Group snapshots by any combination of host,label,paths,tags to find a suitable parent [default: "host,label,paths"]
```
nardoor, line 162: It seems to me that `find a suitable parent` has meaning for the `backup` command (find a parent snapshot to do relative backup)

But as a global option, it won't always have the `find a suitable parent` meaning. For the `find` command or `snapshot` command I don't think it makes sense.

What you

src/commands/find.rs
```diff
@@ -36,15 +36,6 @@ pub(crate) struct FindCmd {
     #[clap(value_name = "ID")]
     ids: Vec<String>,
 
-    /// Group snapshots by any combination of host,label,paths,tags
```
nardoor, line 39: I see that you removed the `group_by` option in `find.rs` and `snapshots.rs` and use the global one.

But for `forget.rs` I see that you keep a command specific `group_by` option.

Could you tell me about the reason for that?

src/config.rs
aawsome: Ah, thanks. This was by copy&paste and I wanted to change it, but forget that. Is fixed.

src/commands/find.rs
aawsome: Yes, for `forget` and `backup` (via `ParentOptions`) there are still command specific `group_by` options.

Those were already present in the config profiles. I kept them for 2 reasons:
- removing would be a breaking change, i.e. users may need to adapt their configs
- I think there might be use cases were users want different standard groupings (e.g. for `snapshots`) than for `backup` and especially for `forget` (and set this in the config profile). So I kept the command-specific options but make them using the global one if not given. 

For `find` and `snapshots` the `group_by` was a CLI-only option. Using the CLI `--group-by` option still works with the global option in the same way.

nardoor:
lgtm!

nardoor (approved):

Issue #426, opened 2023-02-14
Rustic cannot read password from file without newline


waspoza:
I have password file that doesn't end with 0xA. It works well with restic, but rustic throws error:
```
$ ./rustic -p ~/rpw.txt -r ~/mnt/nfs/rustic init
using no config file (./rustic.toml doesn't exist)
Error: error reading password file

Caused by:
    unexpected end of file
```
Would be nice to have same file usable by both programs.

aawsome:
Thanks for opening this issue @waspoza !
You are right, a password file without `\n` doesn't work. This is actually a feature (extra check) of https://github.com/conradkleinespel/rpassword which I'm using to read the password.

As a workaround you could simply add a newline to your password file which would then work with restic as well as with rustic.
Or, you could put your password in the config file using `password=...`

aawsome:
Is anyone willing to add a handling of the "unexpected end of file" error?

waspoza:
Okay, since rustic is using a config file, ill just put the password there. Thanks!

fa993:
> Is anyone willing to add a handling of the "unexpected end of file" error?

@aawsome I'd be happy to do this, but what would the expected error handling look like?

Sathvik1007:
@aawsome 
I am new to open source and I searched but I couldn't find the file so can you tell me the file where the password is read .

aawsome:
@Sathvik1007 I was talking about the config file whis is usually stored in $home/.config/rustic. See the examples folder for example config files.

aawsome:
@fa993 oh, you are right, in case oft an error you can't simply take the password even without newline - the 'Err' doesn't conatain this information. I think the best is to open a feature request to the rpassword crate...

fa993:
@aawsome Hey, I went through the docs at rpassword.... 
We use the `read_password_from_bufread` to read password from file.
Now this function was merged into the rpassword repo to answer this issue [https://github.com/conradkleinespel/rpassword/issues/19](url). 
This issue details that the above mentioned function was introduced so that it would be easy to test the crate by creating a mocking utility.
Also this is present in the source for rpassword

> //! Finally, in unit tests, you might want to pass a `Cursor`, which implements `BufRead`. In that
> //! case, you can use `read_password_from_bufread` and `prompt_password_from_bufread`:

Considering all of this, the only intention for that function is a testing utility, I wonder if a feature request to rpassword would be accepted, since we plan to use this function in production.

aawsome:
[...]

PR #1563, opened 2025-09-15
feat(commands): allow to identify snapshots with "latest" or "latest~N" (N >= 0)

nardoor:
Depends on merging rustic-rs/rustic#426


aawsome:
Hi @nardoor! Do you wish to continue working on this? I think we should try to add full support for `latest~N` soon ;-)

nardoor:
Hello @aawsome,
 You're right, now that rustic_core/pull/426 is merged, I'll take a look :)

nardoor:
@aawsome 
As we said two months ago, for now I'm not sure how to implement this for the `cat` command that is very different from other in terms of implementation.

Also I have a doubt for the `Repair` command :

It's a bit weird to me that we can in theory provide filters to the `repair` command (from the global options) but we don't use it. Even though I agree that when we `repair snapshots`, the filters don't make much sense.. What do you think ? 

---
Apart from that, I believe it's ready for review :)

aawsome:
@nardoor Thanks a lot!
For `cat` I suggest to just leave it as-is. This is a very technical command and for showing snapshots to users the `snapshots` command is the preferred way.

About `rewrite`: IMO this is a bug and we should use the same filtering also here - to make things equal and predictable to users. Even though most users may want to call `repair` with all snapshots. This can be fixed in a follow-up PR.

aawsome:
I made a PR fixing the repair command's snapshot selection: #1636

PR #1565, opened 2025-09-16
feat!: Add log-level-logfile, -dryrun and -dependencies options; remove quiet options

aawsome:
Note: The new `log-level-*` options change the default behavior of log levels when using a log file.

Previous behavior: `log-level` gives the log-level used in the config file; only warnings/errors are logged to stderr
New behavior: logging to log-file is determined by `log-level-logfile`; logging to stderr is determined by `log-level`.

Instead of using the quiet options for `backup` and `forget`, users should now fine-tune their logging-options.


aawsome:
see also #1562

Issue #942, opened 2023-11-15
Environment variables in the config files not working on windows 


kuyagic:
Environment variables in the config files is not properly expanded on windows 
I set the following 

```
[backup]
exclude-if-present = [".nobackup", "CACHEDIR.TAG"]
iglob-file = ["C:/ProgramData/rustic/ignore.iglob.list"]
```

this is working.
but both  `%PROGRAMDATA%/rustic/ignore.iglob.list`  and `%PROGRAMDATA%\\rustic\\ignore.iglob.list` not working 
because `%PROGRAMDATA%` is expanded as `C:\ProgramData` but not `C:\\ProgramData` result in the path is `C:\ProgramData/rustic/ignore.iglob.list` and `C:\ProgramData\\rustic\\ignore.iglob.list`  which is an invalid path on windows

aawsome:
I marked this as enhancement wish as currently there is no substitution of variables in the config file. See also #996

aawsome:
This feature is a bit more involved as there are more questions arising:
- Do we only want variable substitution in filenames or elsewhere in the config?
- Do we want to allow substition only within strings?
- Do we want to first parse the TOML, then substitute what is appropriate or first substitute and then parse (which would allow to define TOML structures within env variables)?

simonsan:
There is this library https://crates.io/crates/expand_str, which is an alternative to https://docs.rs/crate/shellexpand/3.1.0 which is useful for tilde expansion.

rdelaage:
Is someone already working on this? If not I can draft a PR as I have a use case.

aawsome:
@rdelaage If you want to work on this, a PR is warmly welcome!

aawsome:
I reopen this as #1577 seems to introduce a regression when reading toml configs with content not supposed to be affected by env variable substitution.

aawsome:
I realized that env variable substitution introduces a major drawback:

If we want to substitute env variables marked by whatever mark (in our case we used `$`), we also need to introduce some kind of escaping to allow users to use the same string without substitution.

And this escaping has two problems:
- it will be in any case a breaking change in the config
- it introduces an additional complexity the users need to be aware of.

So I'll revert the functionality introduced in #1577 for all toml files.

I am thinking about introducing a new option, e.g. `use-profile-templates` which can be used to explicitely specigy toml file to use with env substitution. @rdelaage What do you think about this idea? Is name of such an option ok or do you see a better one?

rdelaage:
[...]

PR #1600, opened 2025-11-10
fix: Substitute environment variables in profiles only if explicitly opted-in

### LINKED ISSUE #1598: Rustic 0.10.1 fatal parse error due to special characters in password.
I have several machines with Rustic running as backup program.
Each have separate s3 buckets with separate passwords.

On one of my machines following the update to 0.10.1 I can no longer run any commands.
It stops with the following error message
`error: rustic-rs fatal error: parse error: No such variable: $m`

When checking the config I noticed that the last 2 characters of the password are the last 2 characters in the error message.

None of my other machines are giving errors.
I have downgraded to 0.10.0 which is not giving me this error.

Part of the config
```
[global]
log-level = "info"
log-file = "/var/opt/rustic/archive.log"

[repository]
repository = "opendal:s3"
password = "<rest of password>W*$m"
```

rdelaage:
Hi!

This PR fix the regression introduced in #1577 by adding an opt-in flag for environment variable substitution in profile files. Users can still use `$` in their file if they enable substitution by escaping them (`\$`).

This PR closes #1598 and #942

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -120,16 +120,17 @@ impl RusticConfig {
 
         if let Some(path) = paths.iter().find(|path| path.exists()) {
             merge_logs.push((Level::Info, format!("using config {}", path.display())));
-            let config_content = subst::substitute(
-                &std::fs::read_to_string(AbsPathBuf::canonicalize(path)?)?,
-                &subst::Env,
-            )
-            .map_err(|e| {
-                abscissa_core::error::context::Context::new(
-                    FrameworkErrorKind::ParseError,
-                    Some(Box::new(e)),
-                )
-            })?;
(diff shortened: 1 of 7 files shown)


src/config.rs
```diff
@@ -120,16 +126,20 @@ impl RusticConfig {
...
                 )
-            })?;
+                .map_err(|e| {
+                    abscissa_core::error::context::Context::new(
+                        FrameworkErrorKind::ParseError,
+                        Some(Box::new(e)),
+                    )
+                })?
+            } else {
+                std::fs::read_to_string(AbsPathBuf::canonicalize(path)?)?
```
aawsome: maybe read this to a now variable and use this in the if and else branch?

aawsome (approved):
LGTM! Thanks a lot for the fast PR @rdelaage

PR #1615, opened 2025-12-04
feat(interactive): show summary in snapshots, ls and diff views

aawsome:
When summaries are computed using the `s` shortcut in the ls and diff view, they are now also displayed.
Moreover a `S` shortcut has been added in the ls, diff and snapshot view to show summary for the currently selected entry (and all selected snapshots in the snapshots view).

In the summary number of files, dirs, total (restore) size are shown as well as blobs, deduplicated size, size on-repo and the compression ratio.

The summary look like this:
image

### DIFF
--- a/src/commands/ls.rs
+++ b/src/commands/ls.rs
@@ -1,9 +1,6 @@
 //! `ls` subcommand
 
-use std::{
-    ops::{Add, AddAssign},
-    path::Path,
-};
+use std::{ops::AddAssign, path::Path};
 
 #[cfg(feature = "tui")]
 use crate::commands::tui;
@@ -12,6 +9,7 @@ use crate::{Application, RUSTIC_APP, repository::CliIndexedRepo, status_err};
 use abscissa_core::{Command, Runnable, Shutdown};
 use anyhow::Result;
 
+use derive_more::Add;
 use rustic_core::{
     LsOptions,
     repofile::{Node, NodeType},
@@ -82,7 +80,7 @@ impl Runnable for LsCmd {
 /// Summary of a ls command
 ///
 /// This struct is used to print a summary of the ls command.
-#[derive(Default, Clone, Copy)]
+#[derive(Default, Clone, Copy, Add)]
 pub struct Summary {
     pub files: usize,
     pub size: u64,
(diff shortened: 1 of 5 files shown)


kapitainsky:
I have tried and it works great. Already have learnt a few things about my repos.

These are very useful metrics I was always missing but thought that it is too hard to get.

To have a full picture would it be possible to include uncompressed data stats as well? Effectively the same as now but uncompressed sizes. 

```
total RepoSize     48.4 GiB  54.1 GiB
exclusive RepoSize 16.3 GiB  22.0 GiB 
shared RepoSize 32.1 GiB  32.1 GiB

total DataSize     100.4 GiB  120.1 GiB
exclusive DataSize 60.3 GiB  80.0 GiB 
shared DataSize 40.1 GiB  40.1 GiB
```


Having both would be perfection.

aawsome:
> To have a full picture would it be possible to include uncompressed data stats as well? Effectively the same as now but uncompressed sizes.

In fact it is not that easy - or not as easy as I think you are meaning it ;-)

Can you exactly define how "exclusive DataSize" and "shared DataSize" can be computed?

The point is, deduplication can already happen considering either only the left or only the right side. And then for me it is not clear how these Sizes should be defined - or if "exclusive DataSize" + "shared DataSize" can be equal to the "restore size" (which is currently displayed as "total Size").

To give an example: if the left side is a single 1MiB file containing only zeros and the right side is a single 100MiB file containing only zeros, then the left side are 2 identical blobs of 512kiB zeros and the right side are 200 identical blobs of 515kiB zeros. In the repo, only a single compressed blob is saved which has just a few bytes in size - which is the shared RepoSize; and there is 0 byte exlusive RepoSize on both sides.
What would be "shared DataSize" in this example? 0 bytes? 512kiB? 1MiB? I think one could find arguments for all 3 values...

aawsome:
@kapitainsky I added a few more statistics. The output now looks like this:
grafik

aawsome:
I'll add this output to the ls view (of course only showing the current tree and no diff stats) and the snapshots view (with the possibility to get statistics about multiple snapshots), but I'll do this in a future PR ;-)

PR #1641, opened 2026-01-14
feat: Add Windows7 as release target

aawsome:
see #1610


aawsome:
closing this as windows 7 does not work with the rustup pipeline we use.

PR #1695, opened 2026-03-05
perf: use Cow Node names

tilpner:
Fixes for rustic-rs/rustic_core#487


aawsome:
is now included in #1698

PR #1323, opened 2024-10-14
fix(deps): downgrade self-update to fix problems finding right target

simonsan:
Fixes #1291

Related: https://github.com/jaemk/self_update/issues/136

Issue #1291, opened 2024-10-04
rustic self-update on musl-version tries to install gnu-version


aawsome:
```
rustic self-update
Checking target-arch... x86_64-unknown-linux-musl
Checking current version... v0.9.0
Checking latest released version... v0.9.1
New release found! v0.9.0 --> v0.9.1
New release is compatible

rustic release status:
  * Current exe: "/usr/local/bin/rustic"
  * New exe release: "rustic-v0.9.1-x86_64-unknown-linux-gnu.tar.gz"
  * New exe download url: "https://api.github.com/repos/rustic-rs/rustic/releases/assets/196697744"

The new release will be downloaded/extracted and the existing binary will be replaced.
Do you want to continue? [Y/n] n
```

simonsan:
Opened upsteam issue: https://github.com/jaemk/self_update/issues/139

simonsan:
https://github.com/jaemk/self_update/issues/136

> For anyone else running into this issue, downgrading to self_update@0.39.0 works in the mean time as that was the version before the logic above was introduced.

PR #1699, opened 2026-03-09
chore(deps): unpin and update self_update

GideonBear:
Ref #1323, #1291
`self_update` 0.43.1 is supposed to fix the issue. I advise to still test it, but should be safe to upgrade now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -274,10 +274,10 @@ dependencies = [
  "bytes",
  "form_urlencoded",
  "futures-util",
- "http 1.4.0",
- "http-body 1.0.1",
+ "http",
+ "http-body",
  "http-body-util",
- "hyper 1.9.0",
+ "hyper",
  "hyper-util",
  "itoa",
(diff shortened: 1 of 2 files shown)


aawsome:
Thanks a lot @GideonBear!
Just wondering: Is there a reason why https://github.com/jaemk/self_update/issues/136 is still open?

GideonBear:
> Thanks a lot @GideonBear! Just wondering: Is there a reason why [jaemk/self_update#136](https://github.com/jaemk/self_update/issues/136) is still open?

Yeah, because I want one of my users to verify it's fixed organically (in Topgrade), which needs 2 updates (new version of self_update, and a version to update to). I've just released the second new version of Topgrade since the change, so if you wait a few days I can confirm if it works.

GideonBear:
I've confirmed it to be fixed!

aawsome (approved):
LGTM! Thanks @GideonBear

PR #1704, opened 2026-03-15
fix: Add keep-{within-}hourly to config/full.toml

jullanggit:
These seem to be missing from the full config example, but do exist in the code and some other examples.
I can't test this right now, but I think it should be correct.


aawsome (approved):
Thanks @jullanggit

PR #1718, opened 2026-03-28
feat: add exclude-if-xattr option

hacker:
This would be a `rustic` counterpart of https://github.com/rustic-rs/rustic_core/pull/491

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4466,7 +4466,7 @@ dependencies = [
 [[package]]
 name = "rustic_backend"
 version = "0.6.1"
-source = "git+https://github.com/rustic-rs/rustic_core#a8861e719df3290715430990a8ec372cd272cedd"
+source = "git+https://github.com/rustic-rs/rustic_core#511b66ed5f56be566b404ca0426102bc873e374c"
 dependencies = [
  "aho-corasick",
  "backon",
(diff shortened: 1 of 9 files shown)


hacker:
I'm not entirely sure what your workflow is, as this PR requires its counterpart in `rustic_core` for tests to pass and so does the `rustic_core` PR — tests in `rustic` would fail if core is merged without this PR.

At any rate, just explaining why I made this PR draft. (and still don't know if it was a good idea:)).

hacker:
Now that you've merged the core counterpart, I'll make this one ready for review, but let me know if you want me to update `.lock` or your bot will take care of that.

aawsome:
> Now that you've merged the core counterpart, I'll make this one ready for review, but let me know if you want me to update `.lock` or your bot will take care of that.

Can you please also update the `Cargo.lock` file - else the unit tests will fail.
(There might be some recent security issues in the pipeline, too. You can either also try to fix them or I will make a separate PR for that. Could be also that these will need dependency updates within rustic_core to get fixed, though)

hacker:
Hey, I did update core, but I'm still confused with the whole dependency system and what depends on what, but I do suspect that security issues are mostly with core… Anyway, I don't trust myself to take care of them yet ;-)

aawsome (approved):
LGTM! Thanks @hacker

PR #1722, opened 2026-04-02
feat: add interactive setup wizard, improve progress bar UI, and rename get_snapots_from_ids to get_snapshots_from_ids

bolyki01:


MagratheanUK (approved):

aawsome:
Hi @bolyki01, thanks you very much for the PR!

This PR is pretty big and contains 3 different topics. Would you mind separating these into 3 different PRs? This makes reviewing much easier and increases the chance your PRs are getting merged soon!

I'll try to start reviewing some parts but I'll probably have to stop somwhere in the middle ;-)

src/commands/backup.rs
```diff
@@ -326,6 +327,30 @@ impl BackupCmd {
...
+            }
+        }
+        for path_str in &self.sources {
+            if *path_str != "-" {
+                let path = Path::new(path_str);
+                if !path.exists() {
+                    warn!("source path does not exist: {}", path_str);
+                } else if !path.is_dir() && !path.is_file() {
+                    warn!("source path is not a regular file or directory: {}", path_str);
+                }
```
aawsome: This is identical to above. Can you refactor by creating a functio or closure for this=

src/commands/backup.rs
```diff
@@ -365,6 +390,8 @@ impl BackupCmd {
             .no_scan(self.no_scan)
             .dry_run(config.global.dry_run);
 
+        let backup_start = Instant::now();
+
```
aawsome: The information about elapsed backup time is already contained in the snapshot metadata. Better use this...

src/commands/backup.rs
```diff
@@ -412,6 +441,18 @@ impl BackupCmd {
...
+        let (hours, remainder) = (secs / 3600, secs % 3600);
+        let (minutes, seconds) = (remainder / 60, remainder % 60);
+        if hours > 0 {
+            info!("backup of {source} done in {}h {}m {}s.", hours, minutes, seconds);
+        } else if minutes > 0 {
+            info!("backup of {source} done in {}m {}s.", minutes, seconds);
+        } else {
+            info!("backup of {source} done in {:.1}s.", backup_elapsed.as_secs_f64());
+        }
+
```
aawsome: I'd prefer using some pretty printing from the already used `jiff` crate here instead of doing the pretty-printing ourselves,

src/commands/init.rs
```diff
@@ -117,7 +129,16 @@ pub(crate) fn init_credentials(credential_opts: &CredentialOptions) -> Result<Cr
             .with_confirmation("confirm password", "passwords do not match")
             .interact()
         {
-            Ok(pass) => Credentials::Password(pass),
+            Ok(pass) => {
+                // Password strength feedback
+                if pass.is_empty() {
+                    warn!("Warning: Empty password. The repository data will still be encrypted,");
+                    warn!("  but anyone with access to the repository can decrypt it.");
+                } else if pass.len() < 8 {
```
aawsome: If we do want to report about password strength, I'd prefer using the crate `password-strength` to obtain the strength. This check would warn for "b38d?.a" but not for "123456789" which is kind of senseless, IMO.

src/commands/restore.rs
```diff
@@ -34,13 +38,31 @@ pub(crate) struct RestoreCmd {
     #[clap(flatten)]
     ls_opts: LsOptions,
 
+    /// Show interactive restore summary and confirmation prompt.
+    /// Defaults to true when running in a terminal, false otherwise.
+    #[clap(short = 'i', long = "interactive")]
```
aawsome, line 43: This option is already used for several commands to jump into a TUI mode. I would not use this for another non-TUI interactive mode here..
(Also adding this to `restore` is a 4th topic and should go also into a separate PR)

src/commands/restore.rs
```diff
@@ -72,39 +105,123 @@ impl RestoreCmd {
...
+            println!();
+            if hours > 0 {
+                println!(
+                    "✓ Restore completed in {}h {}m {}s",
+                    hours, minutes, seconds
+                );
+            } else if minutes > 0 {
+                println!("✓ Restore completed in {}m {}s", minutes, seconds);
+            } else {
+                println!("✓ Restore completed in {:.1}s", elapsed.as_secs_f64());
```
aawsome, line 223: see above, please use `jiff`

src/commands/restore.rs
```diff
@@ -72,39 +105,123 @@ impl RestoreCmd {
...
+                );
+            }
         }
+
         if restore_infos.restore_size == 0 {
             info!("all file contents are fine.");
         }
 
+        // Confirmation prompt (unless --yes or non-interactive + piped)
+        if !dry_run && !self.skip_confirm && interactive && restore_infos.restore_size > 0 {
```
aawsome, line 187: I must say I don't like this as for all commands there already is a dry-run option and the default is to do dry-run if this is specified and to run the command without extra confirmation if not.

aawsome:
About the `config` command:
- I wonder if this would better suit into the TUI part, i.e. provide a TUI to do the setup
- Instead of manually writing the `toml`, there are crates like `toml_writer` which should help writing correct toml. (and make this more readable)

PR #1750, opened 2026-05-01
docs: config/README.md: mention string requirement for [repository.options]

jullanggit:
I'm not sure why this is the case, but it seems like non-string keys are not allowed, and produce a somewhat cryptic error message:
```
error: rustic-rs fatal error: parse error: TOML parse error at line 9, column 1
  |
9 | [repository]
  | ^^^^^^^^^^^^
invalid type: boolean `true`, expected a string
```.

This patch mentions this requirement in the readme :)


aawsome:
Thanks @jullanggit! Can you please also adjust the formatting here? Thanks a lot!

jullanggit:
Of course!

aawsome (approved):

PR #1751, opened 2026-05-01
feat: add JSON progress output option

lookfirst:
Introduce a global --json-progress option and RUSTIC_JSON_PROGRESS environment variable that selects a machine-readable progress backend for byte-oriented backup progress.

Emit newline-delimited status objects with elapsed time, remaining time, percent complete, total bytes, and completed bytes, then write a final backup summary object after snapshot creation.

Document the new configuration option and update config/show-config snapshots for the added default.

This was AI supervised as I'm not a rust expert. I tested it and it seemed to work well. It aligns this project more closely with restic in terms of functionality. The changes are pretty minimal, but I'll understand if you don't want this PR or if you want me to iterate on it more.


f18m:
Just my 2cents: as rustic user I think this feature would be awesome to have and would make it much easier to integrate rustic in scripts and/or orchestrators!!

src/commands/backup.rs
```diff
@@ -431,6 +434,56 @@ impl BackupCmd {
...
+    dirs_unmodified: u64,
+    data_blobs: u64,
+    tree_blobs: u64,
+    data_added: u64,
+    data_added_packed: u64,
+    total_files_processed: u64,
+    total_bytes_processed: u64,
+    total_duration: f64,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    snapshot_id: Option<&'a SnapshotId>,
```
aawsome: I would directly use a `SnapshotId` without reference here. Then the struct can be defined without liefetime. Using a reference has minimal memory benefits - but JsonProgressSummary is only created once at a time, so this 24 bytes aren't worth the extra complexity, IMO.

src/commands/backup.rs
```diff
@@ -431,6 +434,56 @@ impl BackupCmd {
...
+}
+
+impl<'a> JsonProgressSummary<'a> {
+    fn new(summary: &'a SnapshotSummary, snapshot_id: &'a SnapshotId) -> Self {
```
aawsome: This `new` is only called at one place. Can you remove it and directly construct the struct below?

aawsome (commented):
Thanks a lot @lookfirst!
Looks good, I have only some nitpicks which I think can be improved.

src/commands/backup.rs
lookfirst: @aawsome Thanks for the feedback. I'm not a rust engineer at all. If you want improvements, you're going to have to do it. I'm super sorry about this. I'm sure it is frustrating to have a PR in this way. Feel free to take the code and do whatever you want with it, or even close this out. I just figured it would be helpful for someone else.

aawsome: Thanks for the feedback @lookfirst! I changed this in your branch.

src/commands/backup.rs
aawsome: is changed.

aawsome (approved):
Thanks a lot @lookfirst

lookfirst:
@aawsome Thank you for your understanding and doing the updates.

PR #1753, opened 2026-05-05
fix(scripts): portable bash shebang and prevented array expansions

Turmaxx:
- Replaced `#!/bin/bash` to `#!/usr/bin/env bash` to improve portability on different operating systems.
- Wrapped $@ with quotes to prevent array expansions.


aawsome (approved):
LGTM

PR #1766, opened 2026-05-31
fix: Throttle configuration example

Valentin-Metz:
Adjusted the example throttle config.
Opendal will refuse to push packets greater than burst size, so it must at least be as large as repository pack size.


aawsome (approved):
Thanks and good catch @Valentin-Metz 
LGTM!

PR #1775, opened 2026-06-04
Add pCloud service example config

mro68:
Adds a pCloud example config under config/services, using the same style as the other service examples.

The example uses placeholder credentials and documents the required pCloud endpoint/root settings.


aawsome (approved):
LGTM! Thanks a lot @mro68

PR #1779, opened 2026-06-08
fix: add CA certificates to Docker image

### LINKED ISSUE #1777: Official Docker image lacks CA certificates, causing HTTPS/S3 backends to panic
## Environment

- Image: `ghcr.io/rustic-rs/rustic:latest`
- rustic version: `0.11.3`
- Platform: Linux x86_64
- Runtime: Docker Compose
- Backend: remote backup repository over HTTPS/S3

## What happened

Running `rustic backup` inside the official Docker image fails during HTTP client initialization:

```text
[INFO] using config /etc/rustic/default.toml
name = "rustic-rs"
operating_system = "Linux [x86_64]"
crate_version = "0.11.3"
explanation = """
Panic occurred in file '/home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/reqwest-0.13.4/src/async_impl/client.rs' at line 2507
"""
cause = 'Client::new(): reqwest::Error { kind: Builder, source: General("No CA certificates were loaded from the system") }'
method = "Panic"
````

The container exits with status 139.

## Expected behavior

The official Docker image should support HTTPS/S3/REST TLS backends out of the box, or clearly document that users need to mount a CA bundle and set `SSL_CERT_FILE`.

A normal error message would also be better than a panic.

## Why this appears to happen

The current Dockerfile uses `FROM scratch` for the final image and copies `/rustic` plus minimal `/etc_files/` into the runtime image. This appears to leave the image without a CA bundle such as `/etc/ssl/certs/ca-certificates.crt`.

Since rustic/reqwest uses system CA certificates for TLS verification, HTTPS/S3 backends can fail in this image.

### PR DESCRIPTION
**Pull request by alexma233**
## What this changes

Fixes #1777 

This PR adds the Alpine CA certificate bundle to the final Docker image.

The runtime image still uses `scratch`; it now also includes:

```text
/etc/ssl/certs/ca-certificates.crt
````

## Why

The official Docker image is currently based on `scratch`. When using HTTPS/S3/REST-style backends, `reqwest` needs system CA certificates for TLS verification.

Without a CA bundle, rustic can panic during HTTP client initialization:

```text
Client::new(): reqwest::Error { kind: Builder, source: General("No CA certificates were loaded from the system") }
```

This happened with `ghcr.io/rustic-rs/rustic:latest` / rustic `0.11.3` while backing up to an OpenDAL B2 repository.

## Testing

I built and pushed a test image with this change, then used it in my Docker Compose backup setup with an OpenDAL B2 repository.

The backup completed successfully:

```text
[INFO] repository opendal:b2:: password is correct.
[INFO] snapshot 8706869a successfully saved.
backup completed for data artifact=8706869a
step backup succeeded
backup task finished successfully
```

This confirms that TLS initialization works when the CA bundle is available in the Docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,8 @@
 FROM alpine AS builder
 ARG RUSTIC_VERSION
 ARG TARGETARCH
-RUN if [ "$TARGETARCH" = "amd64" ]; then \
+RUN apk add --no-cache ca-certificates && \
+    if [ "$TARGETARCH" = "amd64" ]; then \
         ASSET="rustic-${RUSTIC_VERSION}-x86_64-unknown-linux-musl.tar.gz";\
     elif [ "$TARGETARCH" = "arm64" ]; then \
         ASSET="rustic-${RUSTIC_VERSION}-aarch64-unknown-linux-musl.tar.gz"; \
@@ -15,4 +16,5 @@ RUN if [ "$TARGETARCH" = "amd64" ]; then \
 FROM scratch
 COPY --from=builder /rustic /
 COPY --from=builder /etc_files/ /etc/
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 ENTRYPOINT ["/rustic"]


alexma233:
I tested multiple official image versions. They all use the `scratch` runtime image and do not contain a CA bundle, but the behavior changed in `v0.11.3`:

- v0.11.0 official image without CA bundle: OK
- v0.11.2 official image without CA bundle: OK
- v0.11.3 official image without CA bundle: FAIL
- v0.11.3 with this Dockerfile change: OK

The failing version panics during reqwest client initialization:

```text
Client::new(): reqwest::Error { kind: Builder, source: General("No CA certificates were loaded from the system") }
```

I do not know which dependency change exposed the issue, but adding the standard CA bundle to the Docker image fixes the image-level problem and makes HTTPS/TLS backends work consistently.

aawsome (approved):
LGTM! Thanks a lot @alexma233

aawsome:
In 0.11.3 the `reqwest` crate was updated to version 0.13. With this update, the certificates handling changed. Before we had certificates bundled within the binary, now we use the certificates from the OS.

PR #1812, opened 2026-07-26
feat: support exclusion on the opendal source

hacker:
it's a counterpart of the commit in the `rustic_core` implementing support for exclusions on the opendal source


aawsome (approved):
I updated the rustic_core dependency and the MSRV.
LGTM!

PR #1830, opened 2026-08-01
fix: keep completions independent from profiles

### LINKED ISSUE #1503: `rustic completions` shouldn't try to read profile
Let image that I'm building rustic on a machine which has `/etc/rustic/rustic.toml` with permission `0640` and build user hasn't got access to it.

When it runs `rustic completions`... it fails, because it hasn't got access to `/etc/rustic/rustic.toml` .

Does `completions` really need acces to it?

magrathean-uk:
## Summary

Keeps shell completion generation independent from profile loading and adds a regression test with an intentionally invalid profile.

## Why

Users should be able to generate completions even when a configured profile is unavailable or malformed. Completion generation only needs the command definition and should not initialize the runtime configuration.

## Validation

- `cargo test --locked --test completions completions_do_not_load_profiles`
- `git diff --check`

Fixes #1503.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -229,6 +229,12 @@ impl Configurable<RusticConfig> for EntryPoint {
         // That's why it says `_config`, because it's not read at all and therefore not needed.
         let mut config = self.config.clone();
 
+        // Completion generation only needs the command definition. In particular, it must not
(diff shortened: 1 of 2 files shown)


aawsome (approved):
LGTM! Thanks @magrathean-uk

Issue #1469, opened 2025-05-31
Confusing order of `as-path` remapping and path filtering


tilpner:
The order in which the `as-path` remapping and the path exclusion happen leads to a potentially confusing interaction (or I'm holding it wrong). I didn't start with a PR because it might be intended behavior, and it looks like moving the remapping into the sources would be a significant change (and probably involve a PR to `ignore`).

## Example

**Setup**

```sh
$ mkdir -p foo/quux
$ touch foo/quux/{3,4}
$ find foo
foo
foo/quux
foo/quux/4
foo/quux/3
$ rustic init --repo local:repo --password foobar
```

Now I'd like to exclude `quux/4`, which works with `--glob '!foo/quux/4'`:

```sh
$ rustic backup --repo local:repo --password foobar --dry-run --log-level debug --glob '!foo/quux/4' foo |& grep quux/4
[DEBUG] (41) ignore::walk: ignoring foo/quux/4: Ignore(IgnoreMatch(Override(Glob(Matched(Glob { from: None, original: "!foo/quux/4", actual: "foo/quux/4", is_whitelist: true, is_only_dir: false })))))
```

But there's no need to have `foo` as a prefix to all my backup paths, and it'd be more readable if my exclude rules didn't have to repeat it every line, so I trim/replace the prefix with `--as-path /`:

```sh
$ rustic backup --repo local:repo --password foobar --dry-run --log-level debug --as-path / --glob '!quux/4' foo |& grep quux/4
[DEBUG] (1) rustic_core::archiver::tree_archiver: new       file: "/quux/4"

$ rustic backup --repo local:repo --password foobar --dry-run --log-level debug --as-path / --glob '!foo/quux/4' foo |& grep quux/4
[DEBUG] (41) ignore::walk: ignoring foo/quux/4: Ignore(IgnoreMatch(Override(Glob(Matched(Glob { from: None, original: "!foo/quux/4", actual: "foo/quux/4", is_whitelist: true, is_only_dir: false })))))
```

The exclude lists operate on the source path, before the as-path remapping, which can be confusing, as the backup lists the remapped paths, against which the exclude rules should have rejected `quux/4`:

```sh
$ rustic backup --repo local:repo --password foobar --as-path / --glob '!quux/4' foo                  
[INFO] starting to backup foo ...
snapshot ac715191 successfully saved.
[INFO] backup of foo done.

$ rustic --repo local:repo --password foobar ls ac715191
quux
quux/3
quux/4
```

[...]

PR #1841, opened 2026-08-01
docs: clarify as-path filtering order

magrathean-uk:
## Summary

Documents the current filtering order: include/exclude globs are evaluated against the original source path before `--as-path` remapping.

## Status

This is intentionally a draft pending maintainer confirmation that this behavior is the desired contract. The documentation describes what rustic currently does; it does not claim that the interaction is intentional or ideal. If maintainers want filters to operate on the remapped path, this should become a behavior change instead.

## Validation

- Reproduced the ordering with the example from #1469
- `git diff --check`

Related to #1469.


tilpner:
Hi, this documents the existing behavior, but in #1469 I was partially asking "Is this intended behavior?" so I could change it if it wasn't. Documenting this without having a maintainer confirm that this interaction is intended feels a bit premature.

I think it's at least possible for this to have gone unnoticed because the failure mode is just that your backup is a bit less efficient. You might never notice that your excludes aren't applying.

magrathean-uk:
Thanks, that’s a good point. I agree the documentation should not imply that this surprising interaction is intentional before a maintainer confirms the desired contract.

I’ve updated the draft PR to describe the current behavior without claiming to fix #1469. If the intended behavior is for filtering to use the remapped path, I’ll treat that as an implementation change instead of locking in the current order through documentation.

PR #1842, opened 2026-08-01
docs: clarify snapshot glob replacement

### LINKED ISSUE #1647: Cannot combine `backup.globs` and snapshot-specific globs
Here is an example config:

```toml
[backup]
globs = ["!.DS_Store", "!._*", "!.DocumentRevisions-*", "!.Spotlight-*", "!.TemporaryItems", "!.Trashes", "!.fseventsd"]

[[backup.snapshots]]
name = "Dotfiles"
sources = ["/Users/me/.config"]
globs = ["!*.sock", "!history.txt"]
```

When backing up files from `~/.config`, the `backup.globs` config is ignored, because it's overridden by the `globs` list in the `Dotfiles` snapshot definition. So a file like `~/.config/foo/.DS_Store` _will_ be backed up.

I expected both lists of globs to be considered (i.e. merging the two lists), but it looks like one overrides the other.

I don’t know if that’s a bug. Probably not? It was a bit unintuitive for me, but it kinda makes sense. If the lists were magically merged, then it would be impossible to override `backup.globs` when you actually want to.

The workarounds I’ve found all rely on using different options to define excludes at the global and snapshot-specific level:

- Use `backup.iglobs` for shared rules, and `globs` for specific snapshots.
- Use `backup.glob-files` for shared rules, and `globs` or `iglobs` for specific snapshots.
- Use `backup.globs` for shared rules, and `.backupignore` files in the source folders for specific snapshots.
- etc.

I don't know if there's anything to do here. It was a bit of a gotcha for me, so I thought I'd raise the issue, but I’m okay with closing it if no one has an idea for improvement.

magrathean-uk:
## Summary

Clarifies how snapshot-specific list options, especially `globs`, relate to the shared `[backup]` configuration.

## Why

The previous wording did not say whether lists were combined or replaced. The reference now explains the replacement behavior and shows how to keep shared patterns in `glob-files`.

## Validation

- Reviewed the wording against the existing configuration merge behavior
- `git diff --check`

Fixes #1647.

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -275,7 +275,10 @@ See [Global Metrics labels](#global-metrics-labels-globalmetrics-labels).
 ### Backup Snapshots `[[backup.snapshots]]`
 
 **Note**: All of the backup options mentioned before can also be used as
-snapshot-specific option and then only apply to this snapshot.
+snapshot-specific option and then only apply to this snapshot. A snapshot-level
+list replaces the corresponding `[backup]` list; options such as `globs` are
+not concatenated automatically. Put shared patterns in `glob-files` when both
+shared and snapshot-specific glob rules are needed.
 
 | Attribute | Description                                                                                                              | Default Value | Example Value                                                          |
 | --------- | ------------------------------------------------------------------------------------------------------------------------ | ------------- | ---------------------------------------------------------------------- |


fvsch:
Looks like a good improvement, thanks!

PR #1851, opened 2026-08-01
fix: allow scrolling long read-only TUI text

### LINKED ISSUE #1233: rustic v0.8.1 - TUI browser small files view - no long lines wrapping
Viewing small files directly from TUI is very handy feature but in case if some lines are longer than terminal width they can not be seen nor copied.

IMO lines wrapping would solve it.

magrathean-uk:
Fixes #1233

Makes long read-only TUI text horizontally scrollable instead of truncating content that exceeds the available width.

Validation: cargo fmt --check; focused TUI tests; git diff --check.


aawsome (approved):
LGTM! Thanks @magrathean-uk

PR #1862, opened 2026-08-01
feat: support source-specific backup log files

### LINKED ISSUE #1434: Specific log files
Please add a support to use log files specific to each `[[backup.snapshots]]`

magrathean-uk:
Fixes #1434

Allows a configured backup snapshot to select its own log file and reapplies the global logger settings for each source so the destination does not leak into the next backup.

Validation: cargo fmt; cargo check --locked.

### DIFF
--- a/src/commands/backup.rs
+++ b/src/commands/backup.rs
@@ -144,6 +144,11 @@ pub struct BackupCmd {
     #[clap(skip)]
     hooks: Hooks,
 
+    /// Write logs for this backup to a profile-selected file.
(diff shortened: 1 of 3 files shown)


aawsome:
Thanks a lot for the PR @magrathean-uk  I think the unit tests will not work as this new option is not reflected there.

magrathean-uk:
Good catch! I've updated the insta config snapshots to include the new `log_file` field. Unit tests are all passing cleanly now.

